PR #121, opened 2013-01-16
Preserve specified time offsets from YAML

chrisperelstein:
This change causes parse_time function to return a Ruby Time object with an offset specified in the parsed YAML. I've been playing with Jekyll which leverages Pysch for parsing YAML data in posts and it would be nice to have Time objects extracted from posts respect the offset specified in the YAML front matter. Then, I could either display it with the specified time offset, convert it to utc with the Time::utc, or convert it to local time with Time::getlocal. This seems to me to be a little bit better behavior, but there may be a good reason not to do this that I'm not aware of.

### DIFF
--- a/lib/psych/scalar_scanner.rb
+++ b/lib/psych/scalar_scanner.rb
@@ -143,7 +143,7 @@ module Psych
         offset += ((tz[1] || 0) * 60)
       end
 
-      klass.at((time - offset).to_i, us)
+      klass.new(yy, m, dd, hh, mm, ss+us/1000000, offset)
     end
   end
 end


tenderlove:
@chrisperelstein sorry to take so long (2 years :frowning:) to get back on this.  I just looked at your repro script.  It looks like you're calling `Psych.load` [against a value that Psych didn't generate](https://gist.github.com/chrisperelstein/5978935#file-psych_time_test-rb-L24).  `Time#to_s` doesn't produce a sting that's in the YAML format spec.

If this is still an issue, can you please give me a test for it?

chrisperelstein:
@tenderlove My turn to take so long. I seem to have deleted the gist I was using as an example (sorry!) so I'm going to try to recreate something. Don't remember where the snippet of yaml came from that I used but I think it was something I was manually doing in order to fix a time zone changing issue I noticed when using octopress. The functional punchline of the pull request is to simply preserve time offset when specified regardless of where the time came from. My opinion is that Pysch shouldn't discard original offset info by automatically converting to local time offset, especially since the Time object includes the convenient getlocal method.

PR #126, opened 2013-02-14
Add anchor order test.

ingydotnet:
Anchors should be emitted in ascending order.


hsbt:
Closing, since anchor numbering does not change what loads back and I'd rather not change dump output for everyone over it.

PR #148, opened 2013-06-24
Add native jruby support to psych. Fixes #145

### LINKED ISSUE #145: Incorporate JRuby bits into gem proper
For JRuby 1.7.5, I am attempting to offload several libraries as default gems, removing the code from our repository. Psych is on the hit list.

JRuby's psych has maintained compatibility up to now, so migrating our code into the Psych repo should not be a big deal. This would also allow Psych updates to be installable by JRuby users.

atambo:
I'm able to run `rake compile` and `rake test` on jruby now and things mostly work. The only change I had to make was to `PsychToRuby.java` because https://github.com/tenderlove/psych/commit/2c644e184192975b261a81f486a04defa3172b3f moved the `path2class` method from `Psych::Visitors` to `Psych::ClassLoader`. However, there are still 20 test failures.


tenderlove:
@headius do you remember where we were with this?  I'd like to do it, but I can't remember the roadblocks we were hitting.

headius:
I think it was mostly that I didn't know how to set up packaging and distribution with the tools that psych is using (rake-compiler, hoe). The library builds and it's almost nearly completely compatible with the C version :-)

We'd want to update this from JRuby master, of course.

tenderlove:
Ok.  I'll give it a whirl today and see what I can find.  If I hit issues, I'll update this ticket so at least it's documented and I won't have to write embarrassing "I can't remember stuff" comments. ;)

jemc:
Ironically enough given our discussion in #216 about instance variables as public/private implementation details, I just ran into this tidbit of MRI-dependent code that may trip up JRuby like it's currently tripping up Rubinius in the test suite:

https://github.com/tenderlove/psych/blob/47adf879591570daebde62895b3712c216710661/ext/psych/psych_yaml_tree.c#L5-L13

I saw this in this PR: https://github.com/tenderlove/psych/pull/148/files#diff-5e38cee2815f08c9932d19acfd91a797R50

Hopefully JRuby happens to use the names 'mesg' and 'backtrace' as internal variables like MRI.   Otherwise, tests for `Psych::Visitors::YAMLTree#visit_Exception` will fail.

Maybe we can find a better way to get the message and backtrace than reaching into MRI hidden ivars that would serve both JRuby and Rubinius?

headius:
We do emit those internal variables for Marshal, but they're not really there...so Psych isn't seeing them and so exceptions probably aren't serializing properly.

I filed #219 for that issue.

hsbt:
@headius We already have native implementation of psych. Can I close this?

/cc @tenderlove

headius:
@hsbt Yes this can be closed. ruby/psych is where the JRuby ext lives now. Now we just need to make sure it always gets released 🙂

PR #178, opened 2014-01-15
Subclasses of string loses encoding

iwiznia:
When subclassing string, the deserialized object loses the encoding.
This is just the test, I really don't know where to fix this. Would be happy to do it if someone points me in the right direction...


hsbt:
Closing this test-only PR. ASCII-only binary strings load back as UTF-8 whether or not they are subclasses, and changing that would alter the loaded value of existing documents.

PR #195, opened 2014-06-14
Better string output

Iazel:
Actually, string that in origin was like
```yaml
a: >
  some
  inline
  content
```

Will be output as:
```yaml
a: |
  some inline content
```

Instead with this edit, will be output accordingly with `line_width` option:
```yaml
# line_width = 100
a: some inline content

# line_width = 11
a: >
  some inline
  content
```

It's not that big of a change, but I think it's nice to have, especially for meticulous people like me :)

### DIFF
--- a/test/psych/test_psych.rb
+++ b/test/psych/test_psych.rb
@@ -57,7 +57,7 @@ class TestPsych < Psych::TestCase
   end
 
   def test_dump_stream
-    things = [22, "foo \n", {}]
+    things = [22, "foo", {}]
     stream = Psych.dump_stream(*things)
     assert_equal things, Psych.load_stream(stream)
   end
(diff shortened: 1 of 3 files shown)


jirutka:
@tenderlove, could you please review this PR and let us know what we should improve to make this merged?

test/psych/test_psych.rb
```diff
@@ -57,7 +57,7 @@ def test_non_existing_class_on_deserialize
   end
 
   def test_dump_stream
-    things = [22, "foo \n", {}]
+    things = [22, "foo", {}]
```
tenderlove: Why does this test change?

jirutka: Probably due to some bug in @Iazel’s implementation that I’ve fixed afterwards — this change is no longer needed.

I’ll send updated PR rebased against current master…

jirutka:
I’ve removed unnecessary change in `test_psych`, squashed it to one commit and rebased to the current master. The new PR is #215, this one can be closed.

PR #215, opened 2015-01-11
Use appropriate style for serialized strings (replaces #195)

jirutka:
This is fixed and squashed #195 updated to the current master.

---

When psych parses YAML with string in the folded style as:

```yaml
a: >
  some
  inline
  content
```

and serializes it back to YAML,
then it renders the string in the literal style instead:

```yaml
a: |
  some inline content
```

This patch fixes this behaviour. When a string doesn't contain newlines (blank line at the end is ignored) and it's not necessary to be quoted, then it will use plain style or folded style according to its length and the `line_width` option.

```yaml
# line_width = 100
a: some inline content

# line_width = 11
a: >
  some inline
  content
```

### DIFF
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -62,13 +62,14 @@ module Psych
 
       def initialize emitter, ss, options
         super()
-        @started  = false
-        @finished = false
-        @emitter  = emitter
-        @st       = Registrar.new
-        @ss       = ss
-        @options  = options
-        @coders   = []
+        @started    = false
+        @finished   = false
+        @emitter    = emitter
+        @st         = Registrar.new
+        @ss         = ss
+        @options    = options
+        @line_width = options[:line_width]
+        @coders     = []
 
         @dispatch_cache = Hash.new do |h,klass|
           method = "visit_#{(klass.name || '').split('::').join('_')}"
@@ -316,7 +317,7 @@ module Psych
           tag   = 'tag:yaml.org,2002:str'
           plain = false
           quote = false
-        elsif line_width != 0 && o.length > line_width
+        elsif @line_width && o.length > @line_width
           style = Nodes::Scalar::FOLDED
           o += "\n" unless o =~ /\n\Z/  # to avoid non-default chomping indicator
         elsif o =~ /^[^[:word:]][^"]*$/
@@ -588,10 +589,6 @@ module Psych
           @emitter.scalar("#{iv.to_s.sub(/^@/, '')}", nil, nil, true, false, Nodes::Scalar::ANY)
           accept target.instance_variable_get(iv)
         end
-      end
-
-      def line_width
-        @line_width ||= (@options[:line_width] || 0)
       end
     end
   end
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -309,19 +309,20 @@ module Psych
           style = Nodes::Scalar::LITERAL
           plain = false
           quote = false
-        elsif o =~ /\n/
+        elsif o =~ /\n[^\Z]/  # match \n except blank line at the end of string
           style = Nodes::Scalar::LITERAL
         elsif o == '<<'
           style = Nodes::Scalar::SINGLE_QUOTED
           tag   = 'tag:yaml.org,2002:str'
           plain = false
           quote = false
+        elsif line_width != 0 && o.length > line_width
+          style = Nodes::Scalar::FOLDED
+          o += "\n" unless o =~ /\n\Z/  # to avoid non-default chomping indicator
         elsif o =~ /^[^[:word:]][^"]*$/
           style = Nodes::Scalar::DOUBLE_QUOTED
-        else
-          unless String === @ss.tokenize(o)
-            style = Nodes::Scalar::SINGLE_QUOTED
-          end
+        elsif not String === @ss.tokenize(o)
+          style = Nodes::Scalar::SINGLE_QUOTED
         end
 
         ivars = find_ivars o
@@ -587,6 +588,10 @@ module Psych
           @emitter.scalar("#{iv.to_s.sub(/^@/, '')}", nil, nil, true, false, Nodes::Scalar::ANY)
           accept target.instance_variable_get(iv)
         end
+      end
+
+      def line_width
+        @line_width ||= (@options[:line_width] || 0)
       end
     end
   end
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -319,7 +319,6 @@ module Psych
           quote = false
         elsif @line_width && o.length > @line_width
           style = Nodes::Scalar::FOLDED
-          o += "\n" unless o =~ /\n\Z/  # to avoid non-default chomping indicator
         elsif o =~ /^[^[:word:]][^"]*$/
           style = Nodes::Scalar::DOUBLE_QUOTED
         elsif not String === @ss.tokenize(o)
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -301,10 +301,9 @@ module Psych
         quote = true
         style = Nodes::Scalar::PLAIN
         tag   = nil
-        str   = o
 
         if binary?(o)
-          str   = [o].pack('m').chomp
+          o     = [o].pack('m').chomp
           tag   = '!binary' # FIXME: change to below when syck is removed
           #tag   = 'tag:yaml.org,2002:binary'
           style = Nodes::Scalar::LITERAL
@@ -333,14 +332,14 @@ module Psych
             plain = false
             quote = false
           end
-          @emitter.scalar str, nil, tag, plain, quote, style
+          @emitter.scalar o, nil, tag, plain, quote, style
         else
           maptag = '!ruby/string'
           maptag << ":#{o.class}" unless o.class == ::String
 
           register o, @emitter.start_mapping(nil, maptag, false, Nodes::Mapping::BLOCK)
           @emitter.scalar 'str', nil, nil, true, false, Nodes::Scalar::ANY
-          @emitter.scalar str, nil, tag, plain, quote, style
+          @emitter.scalar o, nil, tag, plain, quote, style
 
           dump_ivars o
 
--- a/test/psych/test_string.rb
+++ b/test/psych/test_string.rb
@@ -34,6 +34,30 @@ module Psych
       assert_match(/---\s*"/, yaml)
     end
 
+    def test_plain_when_shorten_than_line_width
+      str = "Lorem ipsum"
+      yaml = Psych.dump str, line_width: 12
+      assert_match /---\s*[^>|]+\n/, yaml
+    end
+
+    def test_folded_when_longer_than_line_width_and_no_newlines
+      str = "Lorem ipsum dolor sit amet, consectetur"
+      yaml = Psych.dump str, line_width: 12
+      assert_match /---\s*>\n(.*\n){3}\Z/, yaml
+    end
+
+    def test_folded_when_longer_than_line_width_and_trailing_newline
+      str = "Lorem ipsum dolor sit\n"
+      yaml = Psych.dump str, line_width: 12
+      assert_match /---\s*>\n(.*\n){2}\Z/, yaml
+    end
+
+    def test_literal_when_inner_newline
+      str = "Lorem ipsum\ndolor\n"
+      yaml = Psych.dump str, line_width: 12
+      assert_match /---\s*|\n(.*\n){2}\Z/, yaml
+    end
+
     def test_cycle_x
       str = X.new 'abc'
       assert_cycle str
--- a/test/psych/test_string.rb
+++ b/test/psych/test_string.rb
@@ -30,32 +30,54 @@ module Psych
     end
 
     def test_doublequotes_when_there_is_a_single
-      yaml = Psych.dump "@123'abc"
-      assert_match(/---\s*"/, yaml)
+      str = "@123'abc"
+      yaml = Psych.dump str
+      assert_match /---\s*"/, yaml
+      assert_equal str, Psych.load(yaml)
     end
 
-    def test_plain_when_shorten_than_line_width
+    def test_plain_when_shorten_than_line_width_and_no_final_line_break
       str = "Lorem ipsum"
       yaml = Psych.dump str, line_width: 12
       assert_match /---\s*[^>|]+\n/, yaml
+      assert_equal str, Psych.load(yaml)
     end
 
-    def test_folded_when_longer_than_line_width_and_no_newlines
-      str = "Lorem ipsum dolor sit amet, consectetur"
+    def test_plain_when_shorten_than_line_width_and_with_final_line_break
+      str = "Lorem ipsum\n"
       yaml = Psych.dump str, line_width: 12
-      assert_match /---\s*>\n(.*\n){3}\Z/, yaml
+      assert_match /---\s*[^>|]+\n/, yaml
+      assert_equal str, Psych.load(yaml)
     end
 
-    def test_folded_when_longer_than_line_width_and_trailing_newline
+    def test_folded_when_longer_than_line_width_and_with_final_line_break
       str = "Lorem ipsum dolor sit\n"
       yaml = Psych.dump str, line_width: 12
       assert_match /---\s*>\n(.*\n){2}\Z/, yaml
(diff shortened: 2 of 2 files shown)


lib/psych/visitors/yaml_tree.rb
```diff
@@ -310,19 +310,20 @@ def visit_String o
...
           quote = false
-        elsif o =~ /\n/
-          style = Nodes::Scalar::LITERAL
         elsif o == '<<'
           style = Nodes::Scalar::SINGLE_QUOTED
           tag   = 'tag:yaml.org,2002:str'
           plain = false
           quote = false
+        elsif o =~ /\n[^\Z]/  # match \n except blank line at the end of string
+          style = Nodes::Scalar::LITERAL
```
tenderlove: Can we keep this above the `'<<'` test?  Any string that contains a newline will surely fail the '<<' test, so it would be better if we can keep it early.

lib/psych/visitors/yaml_tree.rb
```diff
@@ -310,19 +310,20 @@ def visit_String o
...
+        elsif o.length > line_width
```
tenderlove: I'm slightly worried about the impact on performance here.  It adds a conditional and a hash lookup that could be cached.  Could you change the initializer to configure this instance with a lambda so that we don't have to call `line_width` for every string?

Something like this:

```ruby
test_lamba = if options[:line_width]
  width = options[:line_width]
  lambda { |o|  o.length > width }
else
  lambda { |_| false }
end

# Then in the elsif

elsif test_lambda.call(o)
```

(But with correct indentation and better variable names)

lib/psych/visitors/yaml_tree.rb
jirutka: Yeah, we can, I just considered it a little cleaner in this order.

tenderlove: Yes, it looks cleaner, but this is an extremely hot code path.  So (unfortunately) order matters.

lib/psych/visitors/yaml_tree.rb
jirutka: This is quite odd… why not to simply cache the value in an instance variable? It will be IMHO even better for performance than calling lambda for every string.

```ruby
def line_width
  @line_width ||= (@options[:line_width] || Float::INFINITY)
end
```

lib/psych/visitors/yaml_tree.rb
jirutka: Okay, I’ll update it.

lib/psych/visitors/yaml_tree.rb
```diff
@@ -310,19 +310,20 @@ def visit_String o
...
+        elsif o =~ /\n[^\Z]/  # match \n except blank line at the end of string
           style = Nodes::Scalar::LITERAL
         elsif o == '<<'
           style = Nodes::Scalar::SINGLE_QUOTED
           tag   = 'tag:yaml.org,2002:str'
           plain = false
           quote = false
+        elsif o.length > line_width
+          style = Nodes::Scalar::FOLDED
+          o << "\n" unless o =~ /\n\Z/  # to avoid non-default chomping indicator
```
tenderlove: Can you dup the string before doing this?  It will mutate the user's objects, and we don't want to do that.  `o = o + "\n"` should be enough.

lib/psych/visitors/yaml_tree.rb
tenderlove: > It will be IMHO even better for performance than calling lambda for every string.

I'm not sure about that.  If you make the change you've proposed above, you'll be doing 2 comparisons, `@line_width ||=`, and `o.length > line_width`, plus an ivar lookup and 2 method calls, `line_width` and `String#length`.  All these will happen in the common case where someone doesn't provide a line_width, as well as when someone provides a line_width.

What I'm proposing would be an ivar lookup and an essentially noop lambda call in the best case, and the same calculations that you have in the worst case.

lib/psych/visitors/yaml_tree.rb
jirutka: This doesn’t work, it breaks `TestString#test_folded_when_longer_than_line_width_and_no_newlines`. The problem is with [this line](https://github.com/jirutka/psych/blob/patch-2/lib/psych/visitors/yaml_tree.rb#L304).

When I assign it to `str` instead of `o`, then tests passes, but I’m not sure if it’s really okay. Why is `o` assigned to `str` and then both `o` and `str` are used later in this method?

tenderlove: >  Why is o assigned to str and then both o and str are used later in this method?

I don't have a good answer for that.  I think we could delete the `str` variable.

tenderlove: Why is the newline being added anyway?  I don't really understand it.  It doesn't seem good because it means that strings won't correctly roundtrip:

```ruby
    def test_line_width_roundtrip
      xs = 'xxx xxx' * 30
      xz = xs.dup # add the dup to avoid side effects
      yaml = Psych.dump xs, line_width: 12
      assert_equal xz, Psych.load(yaml)
    end
```

jirutka: It’s said there — “to avoid non-default chomping indicator.” Try to remove it and run tests, you’ll see.

Where is this test defined?

lib/psych/visitors/yaml_tree.rb
jirutka: Let’s measure it…

```ruby
require 'benchmark'

n = 10_000_000

str = 'Lorem ipsum dolor'
@options_false = {}
@options_true  = {line_width: 80}

def line_width_false
  @line_width_false ||= (@options_false[:line_width] || Float::INFINITY)
end

def line_width_true
  @line_width_true ||= (@options_true[:line_width] || Float::INFINITY)
end

wider_than_line_false = lambda { |_| false }
wider_than_line_true  = lambda { |o|  str.length > line_width_true }

Benchmark.bmbm do |x|
  x.report('variable:false') do
    n.times do
      if str.length > line_width_false
        a = 1
      end
    end
  end
  
  x.report('lambda:false') do
    n.times do
      if wider_than_line_false[str]
        a = 1
      end
    end
  end


  x.report('variable:true') do
    n.times do
      if str.length > line_width_true
        a = 1
      end
    end
  end
  
  x.report('lambda:true') do
    n.times do
      if wider_than_line_true[str]
        a = 1
      end
    end
  end
end

```

```sh
$ ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-darwin13]

$ ruby bench.rb
Rehearsal --------------------------------------------------
variable:false   2.490000   0.010000   2.500000 (  2.496060)
lambda:false     2.400000   0.000000   2.400000 (  2.406696)
variable:true    1.940000   0.010000   1.950000 (  1.938445)
lambda:true      3.430000   0.010000   3.440000 (  3.447116)
---------------------------------------- total: 10.290000sec

                     user     system      total        real
variable:false   2.360000   0.000000   2.360000 (  2.363374)
lambda:false     2.410000   0.000000   2.410000 (  2.415326)
variable:true    1.880000   0.010000   1.890000 (  1.891928)
lambda:true      3.420000   0.000000   3.420000 (  3.419873)
```

jirutka: Let’s measure it…

```ruby
require 'benchmark'

n = 10_000_000

str = 'Lorem ipsum dolor'
@options_false = {}
@options_true  = {line_width: 80}

def line_width_false
  @line_width_false ||= (@options_false[:line_width] || Float::INFINITY)
end

def line_width_true
  @line_width_true ||= (@options_true[:line_width] || Float::INFINITY)
end

wider_than_line_false = lambda { |_| false }
wider_than_line_true  = lambda { |o|  str.length > line_width_true }

Benchmark.bmbm do |x|
  x.report('variable:false') do
    n.times do
      if str.length > line_width_false
        a = 1
      end
    end
  end
  
  x.report('lambda:false') do
    n.times do
      if wider_than_line_false[str]
        a = 1
      end
    end
  end


  x.report('variable:true') do
    n.times do
      if str.length > line_width_true
        a = 1
      end
    end
  end
  
  x.report('lambda:true') do
    n.times do
      if wider_than_line_true[str]
        a = 1
      end
    end
  end
end

```

```sh
$ ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-darwin13]

$ ruby bench.rb
Rehearsal --------------------------------------------------
variable:false   2.490000   0.010000   2.500000 (  2.496060)
lambda:false     2.400000   0.000000   2.400000 (  2.406696)
variable:true    1.940000   0.010000   1.950000 (  1.938445)
lambda:true      3.430000   0.010000   3.440000 (  3.447116)
---------------------------------------- total: 10.290000sec

                     user     system      total        real
variable:false   2.360000   0.000000   2.360000 (  2.363374)
lambda:false     2.410000   0.000000   2.410000 (  2.415326)
variable:true    1.880000   0.010000   1.890000 (  1.891928)
lambda:true      3.420000   0.000000   3.420000 (  3.419873)
```

lib/psych/visitors/yaml_tree.rb
tenderlove: I haven't added this test, I just saw that the PR would make it fail in this case.  Can you add my test to this PR?

lib/psych/visitors/yaml_tree.rb
tenderlove: Can you do the same benchmark with [benchmark-ips](https://github.com/evanphx/benchmark-ips)?  Also note that we could refactor to a simple method call rather than a lambda:

```diff
diff --git a/lib/psych/visitors/yaml_tree.rb b/lib/psych/visitors/yaml_tree.rb
index aa7f10e..98b708d 100644
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -60,6 +60,18 @@ module Psych
         create emitter, ss
       end
 
+      module Empty
+        def self.call o
+          false
+        end
+      end
+
+      class NonEmpty < Struct.new(:len)
+        def call o
+          o.length > len
+        end
+      end
+
       def initialize emitter, ss, options
         super()
         @started  = false
@@ -79,6 +91,10 @@ module Psych
 
           h[klass] = method
         end
+
+        @len_compare = options[:line_width] ?
+                        NonEmpty.new(options[:line_width]) :
+                        Empty
       end
 
       def start encoding = Nodes::Stream::UTF8
@@ -317,7 +333,7 @@ module Psych
           quote = false
         elsif o =~ /\n[^\Z]/  # match \n except blank line at the end of string
           style = Nodes::Scalar::LITERAL
-        elsif o.length > line_width
+        elsif @len_compare.call o
           style = Nodes::Scalar::FOLDED
           o << "\n" unless o =~ /\n\Z/  # to avoid non-default chomping indicator
         elsif o =~ /^[^[:word:]][^"]*$/
@@ -590,10 +606,6 @@ module Psych
           accept target.instance_variable_get(iv)
         end
       end
-
-      def line_width
-        @options[:line_width] || Float::INFINITY
-      end
     end
   end
 end
```

So you may want to add a test that calls a method that just returns `false`.

jirutka: > Can you do the same benchmark with benchmark-ips?

```
      variable:false    78.894k i/100ms
        lambda:false    84.828k i/100ms
       variable:true    85.861k i/100ms
         lambda:true    76.106k i/100ms
-------------------------------------------------
      variable:false      3.947M (± 6.7%) i/s -     19.724M
        lambda:false      3.987M (± 6.3%) i/s -     19.850M
       variable:true      4.568M (± 5.3%) i/s -     22.839M
         lambda:true      2.828M (± 6.1%) i/s -     14.156M
```

The conclusion is the same. When `line_width` is not set, then difference between variable and lambda is NOT statistically significant. When the `line_width` is set, then variable is significantly faster.

> Also note that we could refactor to a simple method call rather than a lambda:

I hope that you’re not serious…

jirutka: I’ve replaced `Float::INFINITY` with 0 and comparison for zero, now both cases are significantly faster than your lambda. :smirk_cat:

```ruby
require 'benchmark/ips'

str = 'Lorem ipsum dolor'
@options_false = {}
@options_true  = {line_width: 80}

def line_width_false
  @line_width_false ||= (@options_false[:line_width] || Float::INFINITY)
end

def line_width_false_zero
  @line_width_false_zero ||= (@options_false[:line_width] || 0)
end

def line_width_true
  @line_width_true ||= (@options_true[:line_width] || Float::INFINITY)
end

def line_width_true_zero
  @line_width_true_zero ||= (@options_true[:line_width] || 0)
end

wider_than_line_false = lambda { |_| false }
wider_than_line_true  = lambda { |o|  str.length > line_width_true }
wider_than_line_true_zero  = lambda { |o|  line_width_true_zero && str.length > line_width_true_zero }

Benchmark.ips do |x|
  x.report('variable:false') do
    if str.length > line_width_false
      a = 1
    end
  end
  
  x.report('variable:false-zero') do
    if line_width_false_zero != 0 && str.length > line_width_false_zero
      a = 1
    end
  end

  x.report('lambda:false') do
    if wider_than_line_false[str]
      a = 1
    end
  end


  x.report('variable:true') do
    if str.length > line_width_true
      a = 1
    end
  end
  
  x.report('variable:true-zero') do
    if line_width_false_zero != 0 && str.length > line_width_true_zero
      a = 1
    end
  end

  x.report('lambda:true') do
    if wider_than_line_true[str]
      a = 1
    end
  end

  x.report('lambda:true-zero') do
    if wider_than_line_true_zero[str]
      a = 1
    end
  end
end
```

```sh
      variable:false    71.546k i/100ms
 variable:false-zero    76.958k i/100ms
        lambda:false    78.478k i/100ms
       variable:true    78.431k i/100ms
  variable:true-zero    77.238k i/100ms
         lambda:true    66.782k i/100ms
    lambda:true-zero    62.260k i/100ms
-------------------------------------------------
      variable:false      3.981M (± 4.7%) i/s -     19.890M
 variable:false-zero      4.627M (± 4.8%) i/s -     23.087M
        lambda:false      3.972M (± 5.7%) i/s -     19.776M
       variable:true      4.520M (± 7.3%) i/s -     22.431M
  variable:true-zero      4.654M (± 4.6%) i/s -     23.249M
         lambda:true      2.839M (± 4.5%) i/s -     14.158M
    lambda:true-zero      2.112M (± 4.4%) i/s -     10.584M
```

lib/psych/visitors/yaml_tree.rb
jirutka: > I don't have a good answer for that. I think we could delete the `str` variable.

Okay, I’ve removed it in 3c27c08

lib/psych/visitors/yaml_tree.rb
jirutka: Oh, why we’re doing this? We can simply assign `@options[:line_width]` to an instance variable during initialization! Then it’ll be just:

```ruby
...
elsif @line_width && o.length > @line_width
...
```

You can’t beat this!
```
        false:preset      7.491M (± 3.8%) i/s -     37.440M
         true:preset      6.295M (± 5.6%) i/s -     31.384M
```

And cleaner code wins again. :smile_cat:

tenderlove:
Everything looks good so far!  Now can you add this test and make it pass?

```ruby
    def test_line_width_roundtrip
      xs = 'xxx xxx' * 30
      xz = xs.dup # add the dup to avoid side effects
      yaml = Psych.dump xs, line_width: 12
      assert_equal xz, Psych.load(yaml)
    end
```

jirutka:
Well, I’ve read the YAML specification again (especially [this part](http://yaml.org/spec/1.2/2009-07-21/spec.html#id2593651)) and you’re right, we should not add the trailing newline to avoid non-default (strip) chomping operator (e.g. `>-` and `|-`) — it’s the correct behaviour.

I’ve fixed it in af668b4c83692ca8556d0aed97a3c542f7176e75 * and also added more test. The behaviour that you’re testing with `test_line_width_roundtrip` is now covered by [test_folded_strip_when_longer_than_line_width_and_no_newlines](https://github.com/jirutka/psych/blob/af668b4c83692ca8556d0aed97a3c542f7176e75/test/psych/test_string.rb#L65).

_* I prefer to not squash it to be more explicit about this intention — it’s actually not the behaviour that most people expects._

tenderlove:
@jirutka cool.  If you're happy with this PR, I'll merge it.  Let me know!

jirutka:
Yeah, if you’re happy with it too, then it’s ready to be merged. :smile_cat:

jirutka:
Will you also merge it into [Ruby upstream](https://github.com/ruby/ruby/)?

tenderlove:
At some point, yes.  :-)

Issue #217, opened 2015-01-15
Marshalling based on instance-variable dumping for primitive types is surprising to the user 


jemc:
Psych currently attempts to use a special-case strategy for marshalling primitive objects when they have instance variables defined within them.

```ruby
require 'psych'; Psych.dump("foo".tap { |s| s.instance_variable_set(:@bar, 88) })
#=> "--- !ruby/string\nstr: foo\nbar: 88\n"
```

Here's an (admittedly contrived) example of the kinds of problems that can happen.
```ruby
require 'psych'
s = "foo"
s.instance_variable_set(:@str, 88)
Psych.load(Psych.dump(s)) #=> 88
```

For a real world example, see issue #216.

I would argue that this behavior is surprising, and should be opt-in only with a flag the user can specify on `Psych.dump`.  Ruby objects of a class that does not correspond directly to a YAML primitive could still be serialized implicitly (including subclasses of a class that corresponds to a primitive, as requested in issue #43).

If we can agree on how to proceed, I'm happy to do the work of crafting a PR.

jemc:
Going through the source to work on the PR here, I've noticed that psych already distinguishes between Array and subclasses of Array.

https://github.com/tenderlove/psych/blob/47adf879591570daebde62895b3712c216710661/lib/psych/visitors/yaml_tree.rb#L423-L431

So hopefully we can agree to treat Strings and Hashes this same way?  I'll craft a PR.

Issue #216, opened 2015-01-15
In Rubinius, version 2.0.9 emits creatively.


mustmodify:
This behavior is present in rbx 2.2 and 2.4 but not 2.1.

psych 2.0.8
---------------
    rbx-2.4.1 :001 > require 'yaml'
    rbx-2.4.1 :002 > Psych.dump({fu: 'bar'})
    => "---\n:fu: bar\n"
 
psych 2.0.9
------------------

    rbx-2.4.1 :001 > require 'yaml'
    rbx-2.4.1 :005 > Psych.dump({fu: 'bar'})
    => "--- !ruby/hash-with-ivars\nelements:\n  :fu: bar\nivars:\n  :@entries: !ruby/object:Rubinius::Tuple {}\n  :@capacity: 16\n  :@size: 1\n  :@max_entries: 12\n  :@state: &1 !ruby/object:Hash::State\n    compare_by_identity: false\n    head: &2 !ruby/object:Hash::Bucket\n      key_hash: 2112643688174279348\n      value: bar\n      link: \n      state: *1\n      key: :fu\n    tail: *2\n  :@mask: 15\n"

References
http://stackoverflow.com/questions/27949877/why-cant-bundle-find-gems-in-production-after-bundle-deployment
https://github.com/rubinius/rubinius/issues/3284

jemc:
Rubinius uses instance variables in core classes like Hash, as they are implemented in Ruby.  Psych is incorrectly assuming here that any instance variables it finds in a Hash are "extra" instead of part of the language implementation.

mustmodify:
https://github.com/tenderlove/psych/blob/47adf879591570daebde62895b3712c216710661/lib/psych/visitors/yaml_tree.rb#L370

jemc:
Looks like this change was made to address this feature request: https://github.com/tenderlove/psych/issues/43

tenderlove:
> Psych is incorrectly assuming here that any instance variables it finds in a Hash are "extra" (and must be serialized) instead of part of the language implementation.

How do I tell the difference between an ivar that a user set, and one that is specific to the implementation?  Should `instance_variables` be returning implementation specific data?

sodabrew:
Interesting, this is analogous to the issue I have in rubinius/rubinius#3268.

headius:
Seems like Hash in Rubinius needs to hide its instance vars, similar to Array and String:

```
[...]

PR #218, opened 2015-01-16
Fix issue #217 and #216 - Only dump ivars for subclasses of Hash and String.

jemc:
Fix issue #217 and #216 - Only dump ivars for subclasses of Hash and String.


tenderlove:
I think this is fine.  Thanks for the PR!

jemc:
Thanks for the quick merge, @tenderlove!

PR #221, opened 2015-01-17
Fix anchor

nobu:


tenderlove:
@nobu I added you to the project, so feel free to commit anytime!

PR #225, opened 2015-02-17
Fix CVE-2014-9130

flori:
See https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2014-9130

This is already fixed in libyaml via this patch:

https://bitbucket.org/xi/libyaml/commits/2b9156756423e967cfd09a61d125d883fca6f4f2


tenderlove:
Thanks!

PR #233, opened 2015-05-03
update minitest gem to 5.x, add Ruby 2.2 test target

miketheman:
Follows #213
Closes #212


miketheman:
Bump for @tenderlove @hsbt - anything I can do here to help this through?

hsbt:
Thanks for your patch. looks good.

Issue #234, opened 2015-05-05
Psych JRuby gem seems to have various build and install issues.


headius:
We've been having trouble getting the Psych gem to build and install consistently.

Sometimes the snakeyaml version number causes a (benign?) error:

```
/Users/headius/projects/psych/lib/psych/versions.rb:2: warning: already initialized constant DEFAULT_SNAKEYAML_VERSION
[INFO] Scanning for projects...
/Users/headius/projects/psych/.tesla.Mavenfile
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project no_group_id_given:psych:0.0.0 (/Users/headius/projects/psych/.tesla.Mavenfile) has 1 error
[ERROR]     'dependencies.dependency.version' for org.yaml:snakeyaml:jar must be a valid version but is '${snakeyaml.version}'.
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
```

@tenderlove provided a log of his [issues, starting from a clean environment](https://gist.github.com/tenderlove/ad4ebdff32e5d0fa2ee8), with a [command log](https://gist.github.com/tenderlove/c0313b017f6cf55a2e61)

I managed to get things to this point by changing around ruby-maven versions in JRuby and having latest ruby-maven installed, but I have no idea which fixes are good and which are bad.

Help us @mkristian...you're our only hope.

headius:
This is still on the `jruby` branch until @tenderlove says otherwise.

mkristian:
quite often I am hopeless (I am told) but I will do my best ;)

the latest ruby-maven does always spawn a new process for running maven. the older one tries to run maven in the same JVM as JRuby. but looking at the Rakefile it is better to set the ```snakeyaml.version``` via the command-line arguments. will make a PR for this.

then I have the feeling that jruby-9k has problems with some yaml data: https://github.com/jruby/jruby/issues/2910
will try to isolate this more.

mkristian:
actually with #235 and latest ruby-maven this problems is solved.

tenderlove:
I merged #235, but I still can't install.  I'm getting the same errors shown in the log that @headius linked to. :(

[...]

PR #235, opened 2015-05-06
little ruby-maven related improvements

mkristian:
* pass in the snakeyaml.version via the commandline
* use the jruby version which runs rake for the java compilation as well


mkristian:
fixes the sporadic ```${snakeyaml.version}``` errors on #234

PR #236, opened 2015-05-27
remove ruby-maven setup and use snakeyaml jar from git repo

mkristian:


mkristian:
actually it would be simple open-uri download from maven central instead of putting snakeyaml.jar into the github repo

PR #245, opened 2015-08-27
Drop Ruby 1.9 support

hsbt:
Ruby 1.9 is already ended at Feb, 23, 2015. 

see. https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/

I think psych should drop to support EOL version of Ruby.

@tenderlove How do you think about this?


hsbt:
I discussed this issue with @tenderlove at RubyConf Taiwan 2015. He accepted this PR.

PR #246, opened 2015-08-27
YAML.dump fails to write a file with UTF-8 string

yyamano:
I got Encoding::UndefinedConversionError when dumping a hash containing Japanese UTF-8 string.to a file. See test code in the change, http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-list/50219 and http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-list/50220 .


hsbt:
@yyamano Thank you for your contiribution. I will merge this change to ruby trunk.

Issue #149, opened 2013-06-25
YAML.load_file returns false for empty files


robinboening:
When loading an empty yml file with YAML.load_file it returns false.
This behavior feels a bit strange to me and I would expect to get an empty Hash instead.

I often notice myself writing code like this and adding a comment to remember why I did this:

```
YAML.load_file(file) || {} if File.exists?(file) # YAML.load_file returns false if file is empty.
```

Is there a certain reason for returning false?
How big would be the impact when changing its behavior?

cheers!
Robin

tuexss:
Like this? https://github.com/tuexss/psych/commit/870fcd76b567a485462c5dd829dd69ffe8dcfdfb

tenderlove:
@tuexss I think adding a default like that would work well, but the patch might need to be a little different. I think currently it will raise an ENOENT exception in the case that the file doesn't exist where your patch will return false.  The case @parkr is talking about is when there is an empty file.

tuexss:
@tenderlove ah, got it. so more like this: https://github.com/tenderlove/psych/commit/f5bad0151d733aff5b4affddaa9630b37ce470f2

I guess I should also add some tests for this

tenderlove:
@tuexss (I think) that won't work either because you could have a non-empty yaml file that legitimately represents `nil` or `false`.

For example:

```
irb(main):001:0> require 'psych'
=> true
irb(main):002:0> File.write 'x.yml', Psych.dump(false)
=> 14
irb(main):003:0> Psych.load_file 'x.yml'
=> false
irb(main):004:0> File.write 'x.yml', Psych.dump(nil)
=> 9
irb(main):005:0> Psych.load_file 'x.yml'
=> nil
irb(main):006:0> File.write 'x.yml', Psych.dump(false)
=> 14
irb(main):007:0> Psych.load_file 'x.yml', Object.new
=> false
irb(main):008:0> File.write 'x.yml', Psych.dump(nil)
=> 9
irb(main):009:0> Psych.load_file 'x.yml', Object.new
=> nil
irb(main):010:0> File.write 'x.yml', ''             
=> 0
irb(main):011:0> Psych.load_file 'x.yml'            
=> false
irb(main):012:0> Psych.load_file 'x.yml', Object.new
=> #<Object:0x007fb4c3002dd0>
irb(main):013:0>
```

I think this patch covers all the bases, but I'm not sure if I really love it:

```diff
diff --git a/lib/psych.rb b/lib/psych.rb
[...]

PR #264, opened 2016-01-08
Add optional fallback return value parameter

tuexss:
As suggested in https://github.com/tenderlove/psych/issues/149

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -228,6 +228,8 @@ module Psych
   # The version of libyaml Psych is using
   LIBYAML_VERSION = Psych.libyaml_version.join '.'
 
+  FALLBACK        = Struct.new :to_ruby # :nodoc:
+
   ###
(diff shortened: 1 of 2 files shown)


tenderlove:
LGTM. You mentioned in #149 that we should be returning `nil` by default. I think we should, but I'm unsure if I could change this return value in a bugfix release.  I guess `nil` is still falsey, and the behavior isn't documented, so I'm not sure.

tuexss:
Yes, I agree. Adding a feature is one thing, changing current behavior a different one. Maybe it should fail with a proper exception.

Do you have any idea what I can do about appveyor failing?

sandstrom:
I just hit this. Getting `nil` in return is what I'd expect, and since both `false` and `nil` are falsy I wouldn't be to worried about changing in a bugfix release.

sandstrom:
Nice! ⛵

PR #270, opened 2016-02-09
Support YAML 1.2 Core Schema

hsbt:
I applied patch and added tests from https://bugs.ruby-lang.org/issues/11988 

@tenderlove Can you confirm this? I'm ok to merge and follow YAML 1.2 optional behavior.


hsbt:
I got approvement from @tenderlove

Issue #156, opened 2013-08-21
"-.".to_yaml


thomasth:
This has probably already been reported, but a quick search didn't find it. Sorry if it has. 

The following all produce an 'ArgumentError: invalid value for Float(): "-"'

"-.".to_yaml
["-."].to_yaml
{:a => "-."}.to_yaml

I've found a work around by reverting to syck but it's a little ugly.

ccutrer:
`Psych.dump("0.E+0")` => `ArgumentError: invalid value for Float()` seems to be a very similar case

PR #276, opened 2016-03-29
fix serialization of a string that looks like a valid YAML float

ccutrer:
refs gh-156

(but isn't a valid ruby float)


ccutrer:
after massaging a bit more, this actually fixes parsing of valid YAML floats that look like this (it's a variant of what is already handled of no fractional piece, but with no exponent)

PR #284, opened 2016-07-05
Skips multiple requires of bigdecimal and date

kjeldahl:
The extra requires one pr. bigdecimal or date in the
YAML being loaded was taking a lot of time when used in
a Rails environment.

For instance it took a file containing ~10.000 big decimals
20 seconds to load, and with this change it takes 2.5 seconds


kjeldahl:
Could I get some feedback on this one?

PR #288, opened 2016-08-23
Close Tempfile

nobu:
Fix Tempfile leak.


tenderlove:
@nobu I think you have commit access.  Feel free to commit.  I'll try to move this over to the Ruby organization soon. 😄

PR #296, opened 2016-11-16
Make Psych::Emitter store the IO object in an instance variable

rhenium:
The IO object given to Psych::Emitter#initialize is saved inside
LibYAML, yaml_emitter_t. This is a problem because Ruby's GC marker
can't find it, and can lead to use after free.

So, store the IO object in an instance variable and fetch from it every
time the write handler is called.

The segmentation fault can be reproduced with the following snippet:

~~~ruby
emitter = Psych::Emitter.new(open("/dev/null", "w"))
GC.start
emitter.start_stream(Psych::Parser::UTF16BE) # make it write something
emitter.end_stream
~~~

### DIFF
--- a/ext/psych/psych_emitter.c
+++ b/ext/psych/psych_emitter.c
@@ -8,6 +8,7 @@
 #endif
 
 VALUE cPsychEmitter;
+static ID id_io;
 static ID id_write;
 static ID id_line_width;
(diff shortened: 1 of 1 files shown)


hsbt:
Great! I confirmed to fix segmentation fault with your patch.

Thank you.

PR #299, opened 2016-11-19
feat(visitor): allow to create custom object (instead of Hash) according to context

kwatch:
(Redirected from https://bugs.ruby-lang.org/issues/12960?next_issue_id=12959 .)

New visitor class to create custom object intead of Hash.
It creates custom object according to context (= last mapping key).

* Enables data access like `ydoc.foo.bar` instead of `ydoc['foo']['bar']`.
* Creates custom object instead of Hash, for example `ydoc['teams'][0].is_a?(Team)` and `ydoc['teams'][0].members[0].is_a?(Member)`.

See document of	CustomClassVisitor class for details.

You may feel that CustomClassVisitor is very experimental. I hope that at least changes of 'to_ruby.rb' will be imported to upstream, because implementation of CustomClassVisitor requires monkey-pathing to 'to_ruby.rb'.

### DIFF
--- a/lib/psych/visitors/custom_class.rb
+++ b/lib/psych/visitors/custom_class.rb
@@ -43,6 +43,41 @@ module Psych
     ##     p team.members[0].name     #=> "Haruhi"
     ##     p team.members[0].gender   #=> "F"
     ##
+    ## Example2:
+    ##
+    ##     ## allows `hash.foo` instead of `hash["foo"]`
+    ##     class MagicHash < Hash
+    ##       def method_missing(method, *args)
+    ##         return super unless args.empty?
+    ##         return self[method.to_s]
+    ##       end
+    ##     end
+    ##     ## create visitor with custom hash class
+    ##     require 'psych'
+    ##     require 'psych/visitors/custom_class'
+    ##     classmap = {'*' => MagicHash}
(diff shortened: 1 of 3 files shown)


tenderlove:
I think I'm OK with the `accept_key` and `accept_value` change but not the addition of a new visitor class.  I used the AST / Visitor pattern specifically so that people could implement (and maintain) their own visitors.

kwatch:
Thanks a lot.

> I think I'm OK with the accept_key and accept_value change but not the addition of a new visitor class.

I see. I'll revert the following commits.

* (#b5d3885)  feat(psych): define 'Psych::Visitors::CustomClassVisitor' class
* (#b81d44e)  feat(psych): support key '*' for default custom class
* (#34883b8)  feat(psych): add test script

And I'll leave the rest commits.

* (#4213ebe)  feat(psych): allow to generate custom Hash object
* (#c9061f6)  feat(psych): allow custom Hash object for unknown tagged mapping
* (#95eab41)  feat(psych): add hook point for mapping key and value
* (#080fd60)  feat(psych): add hook points for mapping-like object
* (#b14d98f)  feat(psych): add hook points for merge ('<<')

Is it OK? Give me your advice.

tenderlove:
@kwatch yes, that makes sense.  Please do it!

PR #304, opened 2016-12-09
Remove deprecated code related syck

hsbt:
I removed deprecated alias, name, method prefixes. Their are related with syck that old yaml serializer for ruby language. 

According to the syck was deprecate after Ruby 2.0 release. It's time to remove it when Ruby 2.4 released.

### DIFF
--- a/lib/psych/core_ext.rb
+++ b/lib/psych/core_ext.rb
@@ -4,18 +4,14 @@ class Object
     Psych.add_tag(url, self)
   end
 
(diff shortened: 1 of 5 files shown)


hsbt:
@tenderlove How do you think that deprecation removing?

I hope to remove `Pysch::DeprecatedMethod`. After that, I will release psych-3.0.0.

tenderlove:
@hsbt I think it's good.  It's been long enough, lets remove it!

PR #326, opened 2017-08-02
Add Psych::Handler#event_location

### LINKED ISSUE #197: Psych should expose start_mark and end_mark to handlers
Writing low-level Psych-based parsers is difficult without access to the `start_mark` and `end_mark` data attached to `libyaml` events. Took a shot at hacking this in, but my Ruby-fu at the C level just isn't strong enough yet.

asterite:
Fixes #197

This adds a new reported event to `Psych::Handler`: `event_location`. This reports precise start/end line/column information.

The line/column information provided by `Psych::Parser#mark` is not very useful because it points to the location **past** the event.

For example, when using `Psych::Parser#mark` we can observe this behavior:

```ruby
require "psych"

class MyHandler < Psych::Handler
  attr_accessor :parser

  def scalar(value, anchor, tag, plain, quoted, style)
    puts "-" * 10
    p value
    puts "line: #{@parser.mark.line}"
    puts "col:  #{@parser.mark.column}"
  end

  def streaming?
    true
  end
end

handler = MyHandler.new
parser = Psych::Parser.new(handler)
handler.parser = parser

source = <<~YAML
foo:
  barbaz: 1

qux: 2
YAML

parser.parse source
```

Output:

```text
----------
"foo"
line: 0
col:  4
----------
"barbaz"
line: 1
col:  9
----------
"1"
line: 3
col:  0
----------
"qux"
line: 3
col:  4
----------
"2"
line: 4
col:  0
```

As you can see, the line/column information is exactly **after** the scalar value. In most cases this is not very terrible regarding the line number: it's correct in "foo" and "barbaz". However you can see that for "1", the reported line number is 3 instead of 1. This is because the parser is positioned at "qux" after parsing the scalar, which is a couple of lines below.

`libyaml` provides [`start_mark` and `end_mark`](https://github.com/yaml/libyaml/blob/660242d6a418f0348c61057ed3052450527b3abf/include/yaml.h#L473-L476) as information of an event, and as far as I can see this is the only information not accessible by Psych.

I initially though about adding this info as an extra argument to `Psych::Handler#scalar`. However, this is not backwards compatible. So, we can add a `Psych::Handler#event_location` with a default empty implementation in `Psych::Handler`. If you upgrade the gem version it will still work (unless someone provided a `event_location` method in a custom handler, which is very unlikely).

Another alternative would be to introduce `Psych::Parser#start_mark` and `Psych::Parser#end_mark` though it's a bit difficult and it would maybe be strange because those are related to each event, not to the parser.

With this PR, we can now do this:

```ruby
gem "psych", "=3.0.0.beta3" # I had to use this locally to test it
require "psych"

class MyHandler < Psych::Handler
  attr_accessor :parser

  def event_location(start_line, start_column, end_line, end_column)
    @start_line   = start_line
    @start_column = start_column
    @end_line     = end_line
    @end_column   = end_column
  end

  def scalar(value, anchor, tag, plain, quoted, style)
    puts "-" * 10
    p value
    puts "start line: #{@start_line}"
    puts "start col:  #{@start_column}"
    puts "end line:   #{@end_line}"
    puts "end col:    #{@end_column}"
  end

  def streaming?
    true
  end
end

handler = MyHandler.new
parser = Psych::Parser.new(handler)
handler.parser = parser

source = <<~YAML
foo:
  barbaz: 1

qux: 2
YAML

parser.parse source
```

Output:

```text
----------
"foo"
start line: 0
start col:  0
end line:   0
end col:    3
----------
"barbaz"
start line: 1
start col:  2
end line:   1
end col:    8
----------
"1"
start line: 1
start col:  10
end line:   1
end col:    11
----------
"qux"
start line: 3
start col:  0
end line:   3
end col:    3
----------
"2"
start line: 3
start col:  5
end line:   3
end col:    6
```

As you can see, the information is now super precise and correct :-)

In `libyaml`, the `start_mark` and `end_mark` event info is also provided for other YAML events, like `start_mapping` and `start_sequence`, but this is always the same as the current `Psych::Parser#mark` value, so there's no point in providing it for other events in Psych. This is also why I decided to call it `scalar_location` and not simply `event_location`.

Of course, all of the API details can be discussed (name, other ways to make it backwards compatible), but I think this will be a valuable addition to Psych.

### DIFF
--- a/ext/java/PsychParser.java
+++ b/ext/java/PsychParser.java
@@ -190,6 +190,12 @@ public class PsychParser extends RubyObject {
             while (true) {
                 event = parser.getEvent();
 
+                IRubyObject start_line = runtime.newFixnum(event.getStartMark().getLine());
+                IRubyObject start_column = runtime.newFixnum(event.getStartMark().getColumn());
+                IRubyObject end_line = runtime.newFixnum(event.getEndMark().getLine());
+                IRubyObject end_column = runtime.newFixnum(event.getEndMark().getColumn());
+                invoke(context, handler, "event_location", start_line, start_column, end_line, end_column);
+
                 // FIXME: Event should expose a getID, so it can be switched
                 if (event.is(ID.StreamStart)) {
                     invoke(context, handler, "start_stream", runtime.newFixnum(YAML_ANY_ENCODING.ordinal()));
@@ -277,6 +283,7 @@ public class PsychParser extends RubyObject {
         
     private void handleScalar(ThreadContext context, ScalarEvent se, boolean tainted, IRubyObject handler) {
         Ruby runtime = context.runtime;
+
         IRubyObject anchor = stringOrNilFor(runtime, se.getAnchor(), tainted);
         IRubyObject tag = stringOrNilFor(runtime, se.getTag(), tainted);
         IRubyObject plain_implicit = runtime.newBoolean(se.getImplicit().canOmitTagInPlainScalar());
--- a/ext/psych/psych_parser.c
+++ b/ext/psych/psych_parser.c
@@ -16,6 +16,7 @@ static ID id_start_sequence;
 static ID id_end_sequence;
 static ID id_start_mapping;
 static ID id_end_mapping;
+static ID id_event_location;
 
 #define PSYCH_TRANSCODE(_str, _yaml_enc, _internal_enc) \
   do { \
@@ -232,6 +233,12 @@ static VALUE protected_end_stream(VALUE handler)
     return rb_funcall(handler, id_end_stream, 0);
 }
 
+static VALUE protected_event_location(VALUE pointer)
+{
+    VALUE *args = (VALUE *)pointer;
+    return rb_funcall3(args[0], id_event_location, 4, args + 1);
+}
+
 /*
  * call-seq:
  *    parser.parse(yaml)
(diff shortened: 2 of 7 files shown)


asterite:
I have no idea why AppVeyor failed, maybe a [problem](http://help.appveyor.com/discussions/problems/4830-stuck-on-appveyor-was-unable-to-build-non-mergeable) with it?

asterite:
**Update**: I decided to introduce `event_location` instead of `scalar_location` and call it before **any** event: `start_mapping`, `start_sequence`, `scalar`, etc. This isn't useful in regular YAML format, but it's useful when embedding JSON inside a YAML, for example:

```yaml
foo: [1, 2, 3]
```

In the above YAML if we want to know where the sequence (array) starts and ends, there's no way to know it unless we invoke `event_location` before `start_sequence` and `end_sequence`.

As explained in #197, this is very useful to report precise location errors when treating a YAML file as something that's "compiled" into another form. In the above example if an array is not expected we can point exactly to the `[` character.

asterite:
**Update**: I pushed another commit where `Psych::TreeBuilder` uses the event location to set it to `Psych::Nodes::Node` instances. I believe this is needed because, like before, a Node had all the information from `libyaml` except correct start/end location. In [stack overflow](https://stackoverflow.com/a/29595013) people are proposing hacks to solve this, but even with these hacks the location info is not correct. This would solve the issue in both the `Handler` and `Node` places.

tenderlove:
Thanks for this work! It's great!

PR #328, opened 2017-09-05
Parse.load with an IO/File which has a non-YAML allowed encoding will…

enebo:
… just

set the encoding to UTF-8 and hope for the best.  This appears to be how
libyaml works.

This issue was noticed in yaml/store because it extends pstore which will
create an IO in read_only mode as:

```ruby
RD_ACCESS = {mode: IO::RDONLY | IO::BINARY, encoding: Encoding::ASCII_8BIT}
```

The data in the test case I was debugging was in fact UTF-8 data and MRI was
happy to take this 8bit IO and pretend it is UTF-8.  Form-fitting ftw.

### DIFF
--- a/ext/java/PsychParser.java
+++ b/ext/java/PsychParser.java
@@ -33,6 +33,8 @@ import java.nio.charset.Charset;
 import java.util.Map;
 
 import org.jcodings.Encoding;
+import org.jcodings.specific.UTF16BEEncoding;
+import org.jcodings.specific.UTF16LEEncoding;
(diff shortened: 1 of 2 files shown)


enebo:
When https://github.com/ruby/psych/pull/328 lands and is released this will be fixed appropriately by psych itself, but I will hack a local fix until then to work around the issue.

enebo:
Found from https://github.com/jruby/jruby/issues/4770

headius:
What's the release status right now?

Issue #281, opened 2016-04-28
Making psych output array in flow format is harder that it should be


hkmaly:
Psych allows to parametrize to_yaml output by defining method encode_with on object. But this parametrization is sparely documented and is not as universal as it could be. For example, to force all arrays to display in FLOW style, following might work:

```
class Array
  def encode_with(coder)
    coder.style = Psych::Nodes::Mapping::FLOW
    coder.tag = nil
    coder.seq = self
  end
end
```

(Obviously, this is just easy to test example: in real use, there would either be some if or it wouldn't be Array directly).

But this simple modification doesn't work, because in psych/visitors/yaml_tree.rb inside function emit_coder the coder.style is ignored for seqences by this lines:

```
        when :seq
          @emitter.start_sequence nil, c.tag, c.tag.nil?, Nodes::Sequence::BLOCK
```

Why couldn't it be 

```
        when :seq
          @emitter.start_sequence nil, c.tag, c.tag.nil?, c.style
```

instead?

tenderlove:
I think it could be. Can you write a a patch?

hkmaly:
You mean like this?
```
diff -Naur 1/yaml_tree.rb 2/yaml_tree.rb
--- 1/yaml_tree.rb      2016-04-28 23:12:40.000000000 +0200
+++ 2/yaml_tree.rb      2016-04-28 23:13:45.000000000 +0200
@@ -568,7 +568,7 @@
         when :scalar
           @emitter.scalar c.scalar, nil, c.tag, c.tag.nil?, false, Nodes::Scalar::ANY
         when :seq
-          @emitter.start_sequence nil, c.tag, c.tag.nil?, Nodes::Sequence::BLOCK
+          @emitter.start_sequence nil, c.tag, c.tag.nil?, c.style
           c.seq.each do |thing|
             accept thing
           end
```

hkmaly:
Actually, while at this, maybe the scalar (two lines above the seq) should be changed as well - but that might require to set some default.

hkmaly:
Any progress on this?

sandstrom:
friendly ping @tenderlove 😄

tenderlove:
Yes, please send a PR. Thanks!

hkmaly:
Ok, I tried.

hkmaly:
@tenderlove I suppose we can close this as you solved it in #329 good enough?

PR #329, opened 2017-09-10
Make possible to set coder output style for arrays and scalars.

hkmaly:
Make possible to set coder output style for arrays and scalars. Well originally I wanted arrays but while I was at it ...

See #281 for discussion.

Ok, I can try making pull request ... it will be my first but how hard can that be?

### DIFF
--- a/lib/psych/coder.rb
+++ b/lib/psych/coder.rb
@@ -66,6 +66,7 @@ module Psych
     # Emit a scalar with +value+
     def scalar= value
       @type   = :scalar
+      @style  = Psych::Nodes::Scalar::ANY if @style == Psych::Nodes::Mapping::BLOCK
       @scalar = value
     end
 
--- a/test/psych/test_coder.rb
+++ b/test/psych/test_coder.rb
@@ -208,6 +208,7 @@ module Psych
       foo = Psych.dump([1,2,3])
       assert_equal foo, "---\n- 1\n- 2\n- 3\n"
     end
+
     class TestArrayFlow < Array
       def encode_with(coder)
         coder.style = Psych::Nodes::Mapping::FLOW
@@ -215,24 +216,29 @@ module Psych
         coder.seq = self
       end
     end
+
     def test_coder_style_array_flow
       foo = Psych.dump(TestArrayFlow.new([1,2,3]))
       assert_equal foo, "--- [1, 2, 3]\n"
     end
+
     def test_coder_style_scalar_default
       foo = Psych.dump('some scalar')
       assert_equal foo, "--- some scalar\n...\n"
     end
+
     class TestStringAny < String
       def encode_with(coder)
         coder.tag = nil
         coder.scalar = self
       end
     end
+
     def test_coder_style_scalar_explicit_default
       foo = Psych.dump(TestStringAny.new('some scalar'))
       assert_equal foo, "--- some scalar\n...\n"
     end
+
     class TestStringQuoted < String
       def encode_with(coder)
         coder.style = Psych::Nodes::Scalar::SINGLE_QUOTED
@@ -240,10 +246,12 @@ module Psych
         coder.scalar = self
       end
     end
+
     def test_coder_style_scalar_quoted
       foo = Psych.dump(TestStringQuoted.new('some scalar'))
       assert_equal foo, "--- ! 'some scalar'\n"
     end
+
     class TestStringFolded < String
       def encode_with(coder)
         coder.style = Psych::Nodes::Scalar::FOLDED
@@ -251,6 +259,7 @@ module Psych
         coder.scalar = self
       end
     end
+
     def test_coder_style_scalar_folded
       foo = Psych.dump(TestStringFolded.new('some scalar'))
       assert_equal foo, "--- ! >-\n  some scalar\n"
--- a/test/psych/test_coder.rb
+++ b/test/psych/test_coder.rb
@@ -203,5 +203,57 @@ module Psych
       assert_equal foo.b, bar.b
       assert_nil bar.c
     end
+
+    def test_coder_style_array_default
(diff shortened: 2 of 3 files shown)


tenderlove:
> Ok, I can try making pull request ... it will be my first but how hard can that be?

Wow!  Thank you for your first pull request. You're doing great so far!  I have two comments, and I'll add them on the diff.

lib/psych/coder.rb
```diff
@@ -66,6 +66,7 @@ def represent_object tag, obj
     # Emit a scalar with +value+
     def scalar= value
       @type   = :scalar
+      @style  = Psych::Nodes::Scalar::ANY if @style == Psych::Nodes::Mapping::BLOCK
```
tenderlove: Is this actually necessary?  If I remove this line, all of the tests including the ones you added still pass.

test/psych/test_coder.rb
```diff
@@ -203,5 +203,57 @@ def test_dump_init_with
...
+      def encode_with(coder)
+        coder.style = Psych::Nodes::Scalar::FOLDED
+        coder.tag = nil
+        coder.scalar = self
+      end
+    end
+    def test_coder_style_scalar_folded
+      foo = Psych.dump(TestStringFolded.new('some scalar'))
+      assert_equal foo, "--- ! >-\n  some scalar\n"
+    end
```
tenderlove: Can you add a newline between methods and classes?

lib/psych/coder.rb
hkmaly: It should be. I'm not completely sure, but I suspect that there are two issues which prevents the code to fail without it.
1) In ext/psych/yaml/emitter.c the YAML_ANY_SCALAR_STYLE is handled exactly same as YAML_PLAIN_SCALAR_STYLE - you just can't force plain style on something which would be invalid with it. Which can hardly be called bug, of course.
2) Psych::Nodes::Scalar::PLAIN is numerically equal to Psych::Nodes::Mapping::BLOCK - which is not bug, but can be seen as arbitrary and just lucky.

It will become very hard to find bug if the condition 2 ever changes, but I'm not sure how to test for it. Condition 1 MIGHT be assumed to be permanent.

test/psych/test_coder.rb
hkmaly: (Where exactly - checking other tests - oh, everywhere) ... of course. Will the pull request update automatically or do I need to do something with it? ... something IS happening ...

lib/psych/coder.rb
hkmaly: Actually, the code will also break if Psych::Nodes::Mapping::BLOCK and Psych::Nodes::Sequence::BLOCK will ever differ. I mean Psych, not just my changes.

tenderlove: Can you write a test that fails if I remove that line?

hkmaly: Not without changing the C part to work worse. However, If you really want to remove that line, I can make test which fails when Psych::Nodes::Mapping::BLOCK, Psych::Nodes::Sequence::BLOCK and Psych::Nodes::Scalar::PLAIN aren't numerically equal, with comments about why that's important (basically, copying the explanation from here).

Or, maybe the line "@style    = Psych::Nodes::Mapping::BLOCK" in Psych::Coder::initialize should be commented with note that without such numerical equivalence the code will break.

singpolyma:
Without this change (which seems stalled some time ago?) currently `encode_with` cannot emit a string-like scalar for values considered special.

tenderlove:
@singpolyma If you can finish this PR, I'm happy to merge.  I'm specifically concerned [here](https://github.com/ruby/psych/pull/329#discussion_r137973605).  Removing that line causes no tests to fail, and I don't understand what it's for.  If we had a test to prove that line is necessary, I'd be happy.

hkmaly:
I explained the issue. There are typedef enums defining constants like YAML_PLAIN_SCALAR_STYLE and YAML_BLOCK_MAPPING_STYLE in C and matching defines like Psych::Nodes::Scalar::PLAIN and Psych::Nodes::Mapping::BLOCK in ruby. As long as you care only about their value, not which one it actually SHOULD be, the line is really superfluous. The moment you starts to care because future change will break the code in weird way, it will become important piece of documentation pointing out one of the places which needs to be paid attention to.

It is not currently possible to write any test which will fail when the line is missing without additional changes in code which would only bring MORE bugs. However, I don't want to remove the line myself because I would feel partially responsible for future breakage.

The fact you, one of developers, are not understanding what the problem is points out that this magic SHOULD be explicitly explained in comments with big warnings about being careful when messing around. Sorry, this is nothing personally about you, just illustration how non-obvious it is.

hkmaly:
There are actually more places where lines like
```
@style  = Psych::Nodes::Sequence::BLOCK if @style == Psych::Nodes::Mapping::BLOCK
```
should be added, but considering both of those constants are 1, not adding them won't cause anything to fail.

hkmaly:
> I believe these changes have been super-ceded by #475

"Super-ceded" wouldn't be the right word. It's just a variant without my concerns about what would happen if the bunch of related constants wouldn't "coincidentally" match.

movermeyer:
@hkmaly Sure. It could be that I'm misreading it, but we're only missing the test [you mentioned that could be added](https://github.com/ruby/psych/pull/329#discussion_r143535006), and not any of the changes currently on this PR?

In which case, this PR could be closed since the primary concern is addressed, and the "safety test" could be added in a separate PR?

(Disclaimer: I'm not a maintainer. I just stumbled across this and thought it was still an active issue, became disheartened, and later surprised when I actually looked at code on `master` and saw that the feature was already implemented by another PR)

hkmaly:
Hmmm ... I think it's clear they don't want to add the test, but yes, no need to keep this open.

PR #333, opened 2017-11-24
Add :symbolize_names option to Psych.load like JSON.parse

k0kubun:
`JSON.parse` has `:symbolize_names` option to symbolize keys of the result. I want its equivalent in `Psych.load`.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -252,9 +252,11 @@ module Psych
   #     ex.file    # => 'file.txt'
   #     ex.message # => "(file.txt): found character that cannot start any token"
   #   end
-  def self.load yaml, filename = nil, fallback = false
+  def self.load yaml, filename = nil, fallback = false, symbolize_names: false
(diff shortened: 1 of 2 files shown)


stomar:
@k0kubun @hsbt

What about the other methods that read yaml, like `::safe_load`, `::load_file`, etc.?

k0kubun:
Ah, that would be nice. Good catch. 

I can send PR but you can realize your idea on your own :+1:

hsbt:
I'm going to release pysch-3.0.0 before Dec 12th. Please submit a pull request at this or next week.

stomar:
@k0kubun I probably won't have the time in the next days to prepare a PR.

k0kubun:
I submitted a patch for `.safe_load` in https://github.com/ruby/psych/pull/337. For `.load_file`, I think we shouldn't add such option for backward compatibility as explained in the PR's description.

PR #337, opened 2017-11-30
Add :symbolize_names option to .safe_load too

k0kubun:
suggested at https://github.com/ruby/psych/pull/333#issuecomment-347689488

## Summary 
I added the same thing as #333 only for `.safe_load`.

## Notes
For `.load_file`, as it may take Hash object in existing arguments, we can't add `:symbolize_names` to it keeping backward compatibility (reason is described below).

<hr>

#### before
```rb
File.write("test.yml", "")
Psych.load_file("test.yml", {}) #=> {}
```

#### after 
After adding `symbolize_names: false` to `.load_file` signature,

```rb
File.write("test.yml", "")
Psych.load_file("test.yml", {}) #=> false
```

`{}` is considered as empty keyword arguments, and second argument of `Psych.load_file` becomes `false`.

### Why adding it to `Psych.load` and `Psych.safe_load` is safe?
That's because `Psych.load` and `Psych.safe_load` don't take Hash object for normal usages.

Unlike `Psych.load_file`, `fallback` argument of `Psych.load` is NOT wrapped with `Psych::FALLBACK`. Thus, if you pass `{}` as `fallback`, `{}.to_ruby` will be called and it will be broken. So this method can't take `{}` (We should pass `Psych::FALLBACK.new({})` if we want to have `{}` as fallback).

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -300,7 +300,7 @@ module Psych
   #
   # A Psych::BadAlias exception will be raised if the yaml contains aliases
   # but the +aliases+ parameter is set to false.
-  def self.safe_load yaml, whitelist_classes = [], whitelist_symbols = [], aliases = false, filename = nil
+  def self.safe_load yaml, whitelist_classes = [], whitelist_symbols = [], aliases = false, filename = nil, symbolize_names: false
     result = parse(yaml, filename)
     return unless result
 
@@ -312,7 +312,9 @@ module Psych
     else
       visitor = Visitors::NoAliasRuby.new scanner, class_loader
     end
-    visitor.accept result
+    result = visitor.accept result
+    symbolize_names!(result) if symbolize_names
+    result
   end
 
   ###
--- a/test/psych/test_psych.rb
+++ b/test/psych/test_psych.rb
@@ -184,20 +184,20 @@ class TestPsych < Psych::TestCase
   end
 
   def test_symbolize_names
-    result = Psych.load(<<-eoyml)
+    yaml = <<-eoyml
 foo:
   bar: baz
 hoge:
   - fuga: piyo
     eoyml
+
+    result = Psych.load(yaml)
     assert_equal result, { "foo" => { "bar" => "baz"}, "hoge" => [{ "fuga" => "piyo" }] }
 
-    result = Psych.load(<<-eoyml, symbolize_names: true)
-foo:
-  bar: baz
-hoge:
-  - fuga: piyo
-    eoyml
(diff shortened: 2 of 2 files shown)


hsbt:
Thanks a lot!

> The reason I didn't add the option to Psych.load_file

Sounds reasonable.

stomar:
Indeed causes the mixing of optional positional arguments and optional keyword arguments problems, see #340.

k0kubun:
> raises an exception

You mean calling {}.to_ruby? It's not handled explicitly and I don't think it's a kind of feature which was maintained.

Anyway, my opinion is here https://github.com/ruby/psych/pull/339#issuecomment-348651743

stomar:
See also #339.

Short version: `load` and `load_file` should work exactly in the same way. `load` blowing up when called with a fallback is a bug. The way the new option is implemented prevents fixing this bug. Furthermore it doesn't make sense that `load` has the new option and `load_file` has not.

k0kubun:
> Short version: load and load_file should work exactly in the same way.

I agree for that point. Obviously it was already an existing problem (wether it's a bug or not depends on intention of the person who implemented. It's not necessarily a bug if the behavior is intentional), but I admit adding an option to only some of them would make the matters worse.

To be clear, my comment in the PR is NOT against the point. I'm in favor of unifying interfaces, but using keyword arguments for future extensions.

stomar:
In the issues related to the fallback option (#149, #264) there was never any hint that `load` should not accept a fallback parameter and blow up instead, it's IMO unimaginable that anyone would implement that kind of behaviour by intention.

The current situation: someone who wants symbolized keys _must_ use `load`, someone who wants to have a custom fallback _must_ use `load_file`. What about someone who needs both? **This API is extremely broken.** (The fact alone that this option can not be added to some methods because of purely "technical" reasons should indicate that something is wrong.)

I'm all for using keyword arguments, but in this case either the new option needs to be implemented differently, or older behavior changed (fallback as keyword argument, if that doesn't cause problems in other methods). I'd rather prefer the latter option, if it's acceptable to break compatibility.

k0kubun:
> someone who wants to have a custom fallback must use load_file

Why do you think you can't use Psych.load(yaml, nil, Psych::FALLBACK.new({})) ? That's the correct usage from previous versions.

stomar:
That's certainly **not** the intended usage,

``` ruby
Psych.load_file("empty.yml", {})
# versus:
Psych.load(yaml, nil, Psych::FALLBACK.new({}))   # ????
```

That would be crazy. Psych::FALLBACK is an implementation detail and not part of the public API.

Issue #340, opened 2017-12-01
The `symbolize_names` keyword argument breaks Psych.load / API conflicts


stomar:
IMO this should be considered before the new option is established in the 3.0 major release.

Current call-seq of Psych.load:

``` ruby
def self.load yaml, filename = nil, fallback = false, symbolize_names: false
  #...
end
```

The mixing of optional arguments and optional keyword argument does not play well together: `fallback` can not be set to a hash (the specifically wished use case when the option was added):

``` ruby
Psych.load("", nil, {})  # => false
```

The provided empty hash is ignored as fallback return value, and the default `false` is returned instead.

The hash is only used when the `symbolize_names` keyword argument is also provided. (However, currently this will cause an exception due to an implementation bug in the handling of the fallback argument; a fix is provided in #339.)

Possible solutions to this problem:

* make (the newly added) symbolize_names an optional positional argument (not very elegant...)
* change all optional arguments to keyword arguments (breaking backwards compatibility...)
* change only the fallback argument to a keyword argument (`fallback: false`); in this case it would be necessary to check whether this might cause similar problems in other methods that already have the fallback option or should have it (e.g. `safe_load` would be a candidate for also having this option)
* ... ?

stomar:
See discussions in #337, #339.

stomar:
Another angle to the current situation:

The API has started to evolve in such a way that various methods that do essentially the same thing (reading yaml) do not provide the same set of options (although these options make sense for all of them). And the crucial problem is that **it will not be possible to implement those options for all methods without API changes**.

That‘s not a good basis for going into a new major release, IMO.


E.g. load has the symbolize_names option while the fallback option is broken and cannot be fixed, load_file has the fallback option but the symbolize_names option cannot be added. And that‘s only for purely „historical“ reasons because the respective PR submitters had interest in only one of the methods.

stomar:
@hsbt @tenderlove

Any thought on this and related discussions?

stomar:
ping?

stomar:
I assume this issue will not be consdered for Psych 3.0 / Ruby 2.5...?

hsbt:
I'm ok with current API implementation. Because Hash and keyword argument will be changed/broke Ruby 2.6 or 3.0. see. https://bugs.ruby-lang.org/issues/14183

We should consider again at Ruby 2.6/3.0 for psych-4.0.0

k0kubun:
[...]

PR #342, opened 2017-12-02
Convert fallback option to a keyword argument

stomar:
Converting the optional fallback argument to a keyword argument fixes a problem that is caused by mixing optional arguments and optional keyword arguments.

Without this change, a hash as fallback value is not handled correctly: in `Psych.load("", nil, {})` the hash is not interpreted as the fallback value, and the default value for the fallback argument is used instead.

This is a possible solution to the problems discussed in #340, #339, and #337.

This would break backwards compatibility, but that might be acceptable in the major version change to 3.0, and moreover I doubt that the fallback argument is widely used.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -259,8 +259,8 @@ module Psych
   #   Psych.load("---\n foo: bar")                         # => {"foo"=>"bar"}
   #   Psych.load("---\n foo: bar", symbolize_names: true)  # => {:foo=>"bar"}
   #
-  def self.load yaml, filename = nil, fallback = false, symbolize_names: false
-    result = parse(yaml, filename, fallback)
+  def self.load yaml, filename = nil, fallback: false, symbolize_names: false
+    result = parse(yaml, filename, fallback: fallback)
     result = result.to_ruby if result
     symbolize_names!(result) if symbolize_names
     result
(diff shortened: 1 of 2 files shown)


stomar:
In case this is adopted, the fallback option could be fixed with a modified version of #339 and the new symbolize_names keyword argument could be kept as it is implemented now, and both options could be made to work with _all_ the related methods (load, load_file, safe_load, etc.).

hsbt:
I prefer this patch for #340 solutions. I will merge and release 3.0.2(and yank 3.0.1)

pda:
How come 3.0.1 was yanked? Doing so is very disruptive to projects already using it.

radar:
Agree (and work) with @pda. We had it locked to 3.0.1 in our `Gemfile.lock` and yanking it broke our CI builds. We needed to do an update to this new version to bring our CI back to green.

The 3.0.1 version did not seem to have any major world-breaking security issues, which (as I understand it) is the _only_ reason you'd use for yanking a version... having it exist alongside a 3.0.2 version would've not caused any harm.

I think in the future you should not yank versions so quickly.

hsbt:
Sorry for your inconvenient experience.

The 3.0.2 broke compatibility of `Psych.load` from 3.0.1 and 3.0.0. So, I hope to avoid confusion of user experience for SemVer semantics. It's my mistake in release management.

pda:
No worries @hsbt, your hard work on Ruby is appreciated :+1:

radar:
Yes it’s very appreciated! Thank you @hsbt for your work.

PR #346, opened 2017-12-19
Fix fallback keyword argument of Psych.load

stomar:
_This replaces #339, adapting it to the switch of the fallback positional argument to a keyword argument;
see discussions there._

There is an error in the implementation of the fallback argument for `Psych.load` which causes an exception; it only works with `Psych.load_file` as expected, currently.

Before the fix:

``` ruby
$LOAD_PATH.unshift("./lib")

require "tempfile"
require "yaml"

Tempfile.create("empty") do |f|
  Psych.load_file(f)                # => false
  Psych.load_file(f, fallback: 42)  # => 42
end

Psych.load("")                      # => false
Psych.load("", nil, fallback: 42)   # => NoMethodError: undefined method `to_ruby' for 42:Integer


# note that Psych.load's fallback can be made to work by wrapping the value like this:
Psych.load("", nil, fallback: Psych::FALLBACK.new(42))  # => 42
# but this strange and unexpected behaviour was never mentioned in any of
# the relevant issues or PRs, nor backed by test cases, nor documented.
# The different behaviour of the option for .load and .load_file clearly
# was not intended but is a bug.
```

After the fix:

``` ruby
Tempfile.create("empty") do |f|
  Psych.load_file(f)                # => false
  Psych.load_file(f, fallback: 42)  # => 42
end

Psych.load("")                      # => false
Psych.load("", nil, fallback: 42)   # => 42
```

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -236,8 +236,9 @@ module Psych
   ###
   # Load +yaml+ in to a Ruby data structure.  If multiple documents are
   # provided, the object contained in the first document will be returned.
-  # +filename+ will be used in the exception message if any exception is raised
-  # while parsing.
+  # +filename+ will be used in the exception message if any exception
+  # is raised while parsing.  If +yaml+ is empty, it returns
+  # the specified +fallback+ return value, which defaults to +false+.
   #
   # Raises a Psych::SyntaxError when a YAML syntax error is detected.
   #
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -260,7 +260,7 @@ module Psych
   #   Psych.load("---\n foo: bar", symbolize_names: true)  # => {:foo=>"bar"}
   #
   def self.load yaml, filename = nil, fallback: false, symbolize_names: false
-    result = parse(yaml, filename, fallback: fallback)
+    result = parse(yaml, filename, fallback: FALLBACK.new(fallback))
     result = result.to_ruby if result
     symbolize_names!(result) if symbolize_names
     result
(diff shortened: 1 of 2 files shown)


stomar:
@tenderlove, @headius, @hsbt

Is the current behavior considered a feature or a bug?

tenderlove:
I'm not sure why it needed to be wrapped with a fallback object in the
first place. I need to figure that out. If it turns out there was no
reason, then I think we should consider this a bug

stomar:
Discussions of the feature happened in #149 and #264.

stomar:
_(rebased and amended with additional test cases)_

sorryeh:
See #358 

Added a fix for fallback, unified the interface & backward compatible with deprecation warnings

🥂

stomar:
@tenderlove

IMO wrapping with a fallback object is not necessary in general. It is only currently necessary because the handling of the fallback value is conveniently done in `parse`, which all of the affected load methods use, and `parse` returns a `Psych::Nodes` object when successful. But this seems to be an implementation detail (and I'm not sure whether the public API of `parse` needs or should have a `fallback` option). The option could just as well be handled in the load methods, without any wrapping necessary. (That's also what #358 does.)

So of course there was a reason for wrapping, but I do not think there was ever any intention of having to explicitly wrap the fallback when using the public API of `load`.

That's why I consider this a bugfix. (On the long run I like the direction @sorryeh has suggested, but that looks like a change for a major release.)

BTW. This PR also adds some more test cases :)

PR #351, opened 2018-03-02
Document Psych.dump options

zverok:
`Psych.dump` options aren't documented anywhere, even despite the fact that there is an explicit reference to those non-existent docs in `Object#to_yaml`:

> See `Psych.dump` for more information on the available `options`.

I hope that I've extracted all possible options and their meaning from code correctly. The one option that intently left undocumented is `:version`, because currently, it seems to accept exactly one "correct" value: `"1.1"`, which is also the default value.

Rendered docs screenshot in default RDOC template:

image

BTW, shouldn't those options be turned into keyword arguments already? `.load` has keyword arguments, which is inconsistent with `.dump`.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -418,6 +418,24 @@ module Psych
   # to control the output format.  If an IO object is passed in, the YAML will
   # be dumped to that IO object.
   #
+  # Currently supported options are:
+  #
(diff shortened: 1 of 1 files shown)


hsbt:
Thank you!

PR #356, opened 2018-04-23
Add more test cases for fallback of Psych.load_file

stomar:
Add test cases for the fallback keyword argument of Psych.load_file to make sure that the fallback is not used for valid YAML that represents nil or false (as mentioned in #149).


stomar:
Since these are only test cases and CI is all green, I'll merge. Hope that's ok.

PR #360, opened 2018-05-23
Add Deprecation Warnings to the unified interface of Psych

sorryeh:
This is a follow up to https://github.com/ruby/psych/pull/358

@tenderlove @yuki24 

# Goal
- Add deprecation warning to the new interface changes

# Follow Up
- Remove the old interface args

_Note_: Once #358 ships, only the `warn` lines will be left

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -265,6 +265,7 @@ module Psych
   #
   def self.load yaml, legacy_filename = NOT_GIVEN, filename: nil, fallback: false, symbolize_names: false
     if legacy_filename != NOT_GIVEN
+      warn 'warning: Passing filename with the 2nd argument of Psych.load is deprecated. Use keyword argument like Psych.load(yaml, filename: ...) instead.'
       filename = legacy_filename
     end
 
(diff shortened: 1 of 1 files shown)


sorryeh:
@tenderlove Added deprecation warnings for the changes in #358.

We can ship 358 to ruby/ruby and ship this slated for a new release.
Do you want me to create another PR for after the deprecation?

🥂

hsbt:
@sorryeh Thanks! It's reasonable changes.

Can you create a new patch for suppressing warnings with test-case?

sorryeh:
@hsbt 
Just to be sure: you mean silencing warnings when test are running, right?

a.k.a adding `$VERBOSE=nil` to `helper.rb`?

otherwise, not sure I follow you properly; thanks for clarifying!

Cheers :)

hsbt:
@sorryeh I got following warnings.

```
# Running:

.......................warning: Passing filename with the 2nd argument of Psych.load is deprecated. Use keyword argument like Psych.load(yaml, filename: ...) instead.
...(snip)
```

We should use the new API with keywords argument on test-suite for silencing them.

jonathanhefner:
> NB: I know @tenderlove wanted to ship this after a little while with #358 merged to ruby. I assume it's because he wants to buffer the release.

@sorryeh  Was this concern ever discussed further?  The current situation is that code which must support Ruby 2.5 and Ruby 2.6 must do the following to avoid deprecation warnings:

```ruby
if Gem::Version.new(Psych::VERSION) >= Gem::Version.new("3.1.0.pre1")
  Psych.safe_load(yaml, permitted_classes: [Date])
else
  Psych.safe_load(yaml, [Date])
end
```

Which is not too pretty.  Additionally, the deprecation warnings can be pretty spammy if you're passing all positional arguments (e.g. `YAML.safe_load(yaml, [], [], false, "path/to/file")`).

It would be nice if the deprecation warnings were disabled until Ruby 2.6 EOL.  At that point, everyone can safely switch to kwargs, unless they need to support obsolete Rubies (in which case, extra workaround code is to be expected).

sorryeh:
@hsbt or @tenderlove I'll defer to either of you for the above question ^, I'm fine disabling the deprecation warnings until 2.6 EOL, curious about your thoughts?

PR #366, opened 2018-07-12
Added tests for special cases and better overall API test coverage

sorryeh:
In relation with/discussed in #358 

Improve test coverage of the API and add Special Case tests

@stomar @tenderlove


hsbt:
Looks good. Thanks!

PR #373, opened 2018-08-24
Try to use version constants from versions.rb again

hsbt:
We removed version constants at fc55b92ed76836ded814149628828bc424bd3af1

But I try to revert it used by `require_relative` again.


hsbt:
@headius When you have any issues with this pull request, Please revert this.

PR #379, opened 2018-11-23
Add uplelvel to deprecation warning of Psych

koic:
This is porting ruby/ruby#1992 to upstream.

This PR adds `uplelvel` to deprecation warning of Psych.
The `uplevel` option was introduced from Ruby 2.5. ruby/psych needs to support Ruby 2.4 or lower.

This PR has `warn_with_uplevel` method emulating `warn 'message', uplevel: 1` in Ruby 2.4 or lower.

And this PR relaxes the warning.
https://github.com/ruby/ruby/pull/1992#discussion_r227214370

## Summary

The deprecation warning log has been added the following commit.
https://github.com/ruby/ruby/commit/1c92766bf0b7394057c00f576fce5464a3037fd9

The following is deprecation warning log change.

### Example code

```console
% cat /tmp/psych_example.rb
require 'psych'

Psych.load("--- foo\n", nil)
```

### Before

```console
% ruby -v
ruby 2.6.0dev (2018-10-21 trunk 65252) [x86_64-darwin17]

% ruby /tmp/psych_example.rb
warning: Passing filename with the 2nd argument of Psych.load is
deprecated. Use keyword argument like Psych.load(yaml, filename: ...)
instead.
```

### After

This patch helps detect argument locations that are deprecated usage.

```console
% cd /path/to/ruby/repo
% make install
% /usr/local/bin/ruby /tmp/psych_example.rb
/tmp/psych_example.rb:3: warning: Passing filename with the 2nd
argument of Psych.load is deprecated. Use keyword argument like
Psych.load(yaml, filename: ...) instead.
```

## Other Information

This log format refers to the deprecation warning of `ERB.new` in Ruby 2.6+.
https://github.com/ruby/ruby/blob/v2_6_0_preview2/lib/erb.rb#L808

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -270,7 +270,7 @@ module Psych
   #
   def self.load yaml, legacy_filename = NOT_GIVEN, filename: nil, fallback: false, symbolize_names: false
     if legacy_filename != NOT_GIVEN
-      warn 'warning: Passing filename with the 2nd argument of Psych.load is deprecated. Use keyword argument like Psych.load(yaml, filename: ...) instead.'
+      warn_with_uplevel 'Passing filename with the 2nd argument of Psych.load is deprecated. Use keyword argument like Psych.load(yaml, filename: ...) instead.', uplevel: 1 if $VERBOSE
       filename = legacy_filename
     end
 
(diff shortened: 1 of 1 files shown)


hsbt:
@koic Thanks! It seems reasonable.

PR #389, opened 2019-02-26
Retired to write CHANGELOG.rdoc

hsbt:
Because It has no update from 2015. The users can retrieve the same information via `git log`.

@tenderlove Do you have any concern about this?


sorryeh:
Well, there is Juanitos' update. Perhaps @tenderlove would rather not have a CHANGELOG what so ever?

tenderlove:
> Do you have any concern about this?

Nope, lets remove it!

PR #400, opened 2019-04-26
Remove string_cache in ScalarScanner

casperisfine:
This is an optional followup to https://github.com/ruby/psych/pull/399 as described in https://github.com/ruby/psych/pull/399#issuecomment-487121564

Because strings used as hash keys are duplicated by MRI, the performance gain brought by `@string_cache` is dubious.

After this patch the allocations look like this:

```
 
allocated objects by file
-----------------------------------
...
    429718  /tmp/bundle/ruby/2.5.0/bundler/gems/psych-a28037330ff1/lib/psych/scalar_scanner.rb


allocated objects by location
-----------------------------------
...
    411049  /tmp/bundle/ruby/2.5.0/bundler/gems/psych-a28037330ff1/lib/psych/scalar_scanner.rb:95
```

So the last remaining (significant) allocations is from `parse_int string.gsub(/[,_]/, '')` and I think they can be eliminated with a small refactoring.

For our application (and according to `memory_profiler`) this patch saves `3.52GB` and `22,546,997` allocations during boot.

I'm not sure how to benchmark this, but I really don't think this cache save mare CPU on parsing than it waste in extra GC cycles, except maybe if there is a lot of repeated scalars inside the parsed documents.

@tenderlove @csfrancis @rafaelfranca @Edouard-chin

I'll try to push another commit to get rid of the parse_int allocations.


casperisfine:
> I'll try to push another commit to get rid of the parse_int allocations.

Done: https://github.com/ruby/psych/pull/400/commits/7d04834b79aa6677b1bff42161311cb79809ed7d

PR #409, opened 2019-07-22
Deduplicate hash keys if they're strings

casperisfine:
As discussed in https://github.com/ruby/ruby/pull/2287#issuecomment-513405834

### Usage

It's fairly common for YAML documents to contain several hashes with similar keys, e.g.:

```yaml
- name: George
  role: admin
- name: Peter
  role: user
```

In such scenarios, the returned datastructure will contain two instances of `"name"` and two instances `"role"`. Assuming the structure will potentially be kept in memory for the program lifetime, it's wasted space. 

### The patch

Since a very long time, MRI automatically freezes strings when they are used as hash keys. e.g. it's equivalent to `key.dup.freeze`.

As such we can apply [`String#-@`](https://ruby-doc.org/core-2.6.3/String.html#method-i-2D-40) to deduplicate it first, without breaking backward compatibility.

### Potential followup

It could be interesting to also deduplicate String values, however since it implies freezing them, it would break backward compatibility. As such I assume it would need to be behind some flag.

@tenderlove @Edouard-chin @rafaelfranca

### DIFF
--- a/test/psych/test_hash.rb
+++ b/test/psych/test_hash.rb
@@ -111,5 +111,19 @@ bar:
 eoyml
       assert_equal({"foo"=>{"hello"=>"world"}, "bar"=>{"hello"=>"world"}}, hash)
     end
+
+    def test_key_deduplication
+      unless String.method_defined?(:-@) && (-("a" * 20)).equal?((-("a" * 20)))
+        skip "This Ruby implementation doesn't support string deduplication"
+      end
+
+      hashes = Psych.load(<<-eoyml)
+---
+- unique_identifier: 1
+- unique_identifier: 2
+eoyml
+
+      assert_same hashes[0].keys.first, hashes[1].keys.first
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


casperisfine:
Quick update, starting from https://github.com/ruby/ruby/commit/e205304a3e9 (MRI 2.5?), string hash keys are automatically deduplicated if and only that string already exist in the fstring table.

So I had to rework the test a bit.

tenderlove:
Can you send this to https://github.com/ruby/psych ?

test/psych/test_hash.rb
```diff
@@ -111,5 +111,19 @@ def test_ref_append
 eoyml
       assert_equal({"foo"=>{"hello"=>"world"}, "bar"=>{"hello"=>"world"}}, hash)
     end
+
+    def test_key_deduplication
+      unless String.method_defined?(:-@) && (-("a" * 20)).equal?((-("a" * 20)))
+        skip "This Ruby implementation doesn't support string deduplication"
+      end
```
tenderlove: Do we need this test in the implementation too?  IOW, will the `to_ruby` visitor break if `:-@` isn't defined?

casperisfine: Hum, somehow I was certain I did put it there, but seems like not. I'll update in a sec.

PR #412, opened 2019-07-24
Update ruby versions

nobu:


hsbt:
Unfortunately, 2.4 to 2.6 definitions do not always point the latest versions with rvm. We need to specify the full version string on Travis.yml.

PR #414, opened 2019-07-24
Freeze parsing option

casperisfine:
As discussed here: https://github.com/ruby/ruby/pull/2287#issuecomment-513865160
Followup: https://github.com/ruby/psych/pull/409

With this option enabled:

  - All the objects returned are frozen.
  - All string instances are deduplicated using `String#-@`.

Note that I had to first refactor `symbolize_names`, because it was mutating the data structures at the end of the parsing phase.

Also, we could go a bit further and deduplicate some specific objects such as empty hashes and arrays, but I opted to keep this PR as simple as possible.

@tenderlove @rafaelfranca @Edouard-chin

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -268,7 +268,7 @@ module Psych
   # YAML documents that are supplied via user input.  Instead, please use the
   # safe_load method.
   #
-  def self.load yaml, legacy_filename = NOT_GIVEN, filename: nil, fallback: false, symbolize_names: false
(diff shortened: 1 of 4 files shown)


tenderlove:
@casperisfine can you rebase this and I'll merge it?

casperisfine:
@tenderlove done!

tenderlove:
@casperisfine thank you!

casperisfine:
You're most welcome!

PR #419, opened 2019-10-18
Remove taint support

jeremyevans:
Ruby 2.7 deprecates taint and it no longer has an effect.
The lack of taint support should not cause a problem in
previous Ruby versions.

I'm not sure if the untaint calls in deduplicate are still needed
after the removal of tainting in the parser.  If they are not
needed, they should be removed.


lib/psych/visitors/to_ruby.rb
```diff
@@ -383,7 +382,8 @@ def deduplicate key
           if key.is_a?(String)
             # Deduplication is not supported by this implementation,
             # but we emulate it's side effects
-            key.untaint.freeze
+            key.untaint if RUBY_VERSION < '2.7'
```
byroot: @jeremyevans this `RUBY_VERSION < '2.7'` can be removed because this version of the method is only executed if `String#-@` doesn't exist (meaning MRI 2.2- IIRC)

lib/psych/visitors/to_ruby.rb
```diff
@@ -371,9 +371,8 @@ def revive_hash hash, o
       if String.method_defined?(:-@)
         def deduplicate key
           if key.is_a?(String)
-            # It is important to untaint the string, otherwise it won't
-            # be deduplicated into and fstring, but simply frozen.
-            -(key.untaint)
+            key.untaint if RUBY_VERSION < '2.7'
```
byroot: Could we make this check "static" (like `if String.method_defined?(:-@)`). `deduplicate` is quite an hot spot, and I'm worried about the performance impact of an extra string comparison.

lib/psych/visitors/to_ruby.rb
jeremyevans: Thanks for the review.  I agree, this can be removed.  Note that the tests don't seem to pass on ruby 2.2 due to the use of String#match? (added in 2.4).

lib/psych/visitors/to_ruby.rb
jeremyevans: Sure, this makes sense.

lib/psych/visitors/to_ruby.rb
byroot: Yeah, I think I broke the compatibility in a later PR, I need to check what we did about that. Either we can remove that conditional from the code, or either I need to remove the `match?` optimisation.

byroot: Answer: we but the minimum version requirement to 2.4: https://github.com/ruby/psych/pull/415

So you can remove the `-@` check.

PR #423, opened 2019-11-24
Fix bug `symbolize_names: true` with null column yaml

kibitan:
Trying to parse YAML which contains `null` column `symbolize_names: true` would raise `NoMethodError` 

### sample code

```bash
$ docker run --entrypoint sh --rm -it ruby:2.7-rc

# ruby -v
ruby 2.7.0preview2 (2019-10-22 master 02aadf1032) [x86_64-linux]
# irb
irb(main):001:0> require 'yaml'
=> true
irb(main):002:0>  Psych.load("null: nullnull", symbolize_names: true)
Traceback (most recent call last):
        8: from /usr/local/bin/irb:23:in `<main>'
        7: from /usr/local/bin/irb:23:in `load'
        6: from /usr/local/lib/ruby/gems/2.7.0/gems/irb-1.1.0.pre.3/exe/irb:11:in `<top (required)>'
        5: from (irb):2
        4: from /usr/local/lib/ruby/2.7.0/psych.rb:276:in `load'
        3: from /usr/local/lib/ruby/2.7.0/psych.rb:606:in `symbolize_names!'
        2: from /usr/local/lib/ruby/2.7.0/psych.rb:606:in `each'
        1: from /usr/local/lib/ruby/2.7.0/psych.rb:607:in `block in symbolize_names!'
NoMethodError (undefined method `to_sym' for nil:NilClass)
```


jeremyevans:
This patch only fixes the `nil` case, it doesn't fix cases where the key is a value other than a String or `nil`.  For example, this is still broken:

```ruby
YAML.load("1: n", :symbolize_names=>true)
```

The patch should be changed to only call `to_sym` if the value is a String, I think.

kibitan:
@jeremyevans thank you for the reply, appreciate.
yes, you are right, I haven't notice about it.  I'm going to fix patch 😊

kibitan:
@jeremyevans I reflect your suggestion by 1271aa7, can you have another look? also, windows build on github action were failed, do you have any idea for it?

jeremyevans:
This looks good to me.  @tenderlove should probably review.

kibitan:
@jeremyevans thank you for your support! 😀

@tenderlove could you have a look when you have time?

jeremyevans:
This no longer applies.  It looks like #480 will fix this.

PR #424, opened 2019-11-28
Update windows.yml

MSP-Greg:
1. Fix run commands to work with PowerShell

2. Change to MSP-Greg/actions-ruby, uses current Rubies

3. Test on ruby-head/master.  'Allow-failure' is not UI friendly on Actions.

Please ping me if ruby-head fails.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,20 +8,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.6.x', '2.5.x', '2.4.x' ]
+        ruby: [ '9.9.x', '2.6.x', '2.5.x', '2.4.x' ]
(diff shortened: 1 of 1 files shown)


MSP-Greg:
I left/forgot about the AppVeyor matrix.  It is:
```yaml
    - ruby_version: "24"
    - ruby_version: "24-x64"
    - ruby_version: "25"
    - ruby_version: "25-x64"
```

The '-x64' versions are tested on GitHub Actions.

1. They could be removed, and leave the 32 bit builds?

2. 32 bit Ruby 2.6 could be added?

GitHub Actions does not have 32 bit MinGW tools, installing them takes a long time.  Hence, to test Windows 32 bit, one must do so on AppVeyor...

MSP-Greg:
Outdated. Closing

PR #432, opened 2020-02-12
Update Actions

MSP-Greg:
1. Remove rvm where possible, add ruby-head, misc
2. Add mingw & mswin Windows testing

In the past, many Ruby 'master/head' builds were not fully tested.  All the builds used are built on GitHub and fully tested.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -5,11 +5,17 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [ 2.7, 2.6, 2.5, 2.4 ]
     steps:
-    - uses: actions/checkout@master
-    - name: Install dependencies
-      run: |
-        gem install bundler --no-document
-        bundle install
-    - name: Run test
-      run: rake
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install dependencies
+        run: bundle install
+      - name: Run test
+        run: rake
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -6,19 +6,18 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        ruby: [ '2.6', '2.5', '2.4', 'ruby-head']
+        ruby: [ ruby-head, 2.7, 2.6, 2.5, 2.4 ]
     steps:
-    - name: Install libraries
-      run: sudo apt install haveged
-    - uses: actions/checkout@master
-    - name: Set up Ruby
-      uses: ruby/setup-ruby@v1
-      with:
-        ruby-version: ${{ matrix.ruby }}
-    - name: Install dependencies
-      run: |
-        gem install bundler --no-document
-        bundle install
-    - name: Run test
-      run: rake
+      - name: Install libraries
+        run: sudo apt install haveged
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install dependencies
+        run: bundle install
+      - name: Run test
+        run: rake
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,20 +8,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.6.x', '2.5.x', '2.4.x' ]
+        ruby: [ mingw, mswin, 2.7, 2.6, 2.5, 2.4 ]
     steps:
-    - uses: actions/checkout@master
-    - name: Set up Ruby
-      uses: actions/setup-ruby@v1
-      with:
-        ruby-version: ${{ matrix.ruby }}
-    - name: Update MSYS2
-      uses: MSP-Greg/msys2-action@master
-      with:
-        base: update
-    - name: Install dependencies
-      run: |
-        gem update --system --no-document --conservative
-        bundle install
-    - name: Run test
-      run: rake
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: MSP-Greg/actions-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          base: update
+      - name: Install dependencies
+        run: bundle install
+      - name: Run test
+        if: matrix.ruby != 'mswin'
+        run: rake
+      - name: Run test
+        if: matrix.ruby == 'mswin'
+        shell: cmd
+        run: |
+          call "%VCVARS%"
+          rake
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,4 @@ deploy: off
 environment:
   matrix:
     - ruby_version: "24"
-    - ruby_version: "24-x64"
     - ruby_version: "25"
-    - ruby_version: "25-x64"


.github/workflows/windows.yml
```diff
@@ -8,20 +8,22 @@ jobs:
...
-        base: update
-    - name: Install dependencies
-      run: |
-        gem update --system --no-document --conservative
-        bundle install
-    - name: Run test
-      run: rake
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: MSP-Greg/actions-ruby@v1
```
hsbt: Why don't you use `ruby/setup-ruby`?

.github/workflows/macos.yml
```diff
@@ -5,11 +5,17 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [ 2.7, 2.4 ]
```
hsbt: Can you add 2.6 and 2.5 to this.

.github/workflows/windows.yml
MSP-Greg: 1. Can't update MSYS2, etc.
2. No mswin build
3. Both mswin & mingw builds are fully tested

.github/workflows/macos.yml
MSP-Greg: I can if you'd like.  Previously, just one job was ran, and it was the default Actions Ruby.

MacOS jobs have a much lower concurrent limit that Ubuntu or Windows jobs, so I thought I'd run low/high...

MSP-Greg: Done

.github/workflows/windows.yml
eregon: Re 1., how about a `setup-msys2` or so action doing just that (and nothing on other OS)? Maybe https://github.com/MSP-Greg/msys2-action already does that?
Re 2., fair point, I wouldn't be against adding the possibility to use a mswin build in `ruby/setup-ruby`. Can you open an issue at https://github.com/ruby/setup-ruby if you think it's important to test on mswin? I know little about Windows, but I thought most people use RubyInstaller (and I'd think extremely few do their own builds).
Re 3., doesn't the RubyInstaller-head build run test-all? If not, can you open an issue for that on https://github.com/oneclick/rubyinstaller2?

I see you used `ruby/setup-ruby` for non-Windows workflows, so I think this change is fine, since anyway there are Windows-specific steps in there. But it would be good to make it available for cases which can share a workflow for all 3 OS.

MSP-Greg: I'm trying to deprecate msys2-action and replace it with actions-ruby, which I could rename setup-ruby.

mswin build - ruby/ruby tests against it.  So, as recently happened with ruby/openssl, mingw passed there, but when they pushed the changes to ruby/ruby, mswin failed.  By having a mswin build available for CI, that doesn't need to happen.

At present, mswin builds are not often used.  But, I know of one company using an embedded Ruby based on RI2, and they're having issues with the fact that their app builds with VS 2019 and the RI2 builds use a different msvc runtime (the one used by MSYS2 tools).  So, there is reason to build/test against mswin/msvc.

> doesn't the RubyInstaller-head build run test-all?

See below.  I'm not sure about spec tests, they may run when it hits RI2, not sure.

https://github.com/oneclick/rubyinstaller2-packages/blob/63699f6b8155435cbb499f60de6e534c107fe9bd/mingw-w64-ruby-head/PKGBUILD#L65-L71

Simply put, my shit works, or, if it occasionally doesn't, it gets fixed quickly.  ruby-loco has been used in CI for almost three years in a lot of repos.  The one time I recall it breaking was over the RubyGems binstub issue, and I added tests for that.  I also actively encourage repos to test with ruby master on all platforms, which benefits ruby/ruby...

eregon: I've given a few more thoughts about this.
I think your 3 points are all valid points to use your action here.

1) is part of package management, I wouldn't want to have that in setup-ruby (https://github.com/ruby/setup-ruby/issues/21)
2) For that to work, one needs a separate `windows` workflow/job, because `mswin, mingw` as Ruby versions just wouldn't make sense on Ubuntu & macOS. If it's a separate workflow/job then it's also easy to simply switch to your action in the windows workflow/job.
3) We should resolve that in RubyInstaller, good catch.

I think it's good to use your action for fine control like testing on both mingw & mswin.

> I'm trying to deprecate msys2-action and replace it with actions-ruby, which I could rename setup-ruby.

I think something with windows, mswin or mingw in the action name would be helpful to clarify it's meant for Ruby on Windows. Another `setup-ruby` would be confusing because it wouldn't be a superset of ruby/setup-ruby.

.github/workflows/ubuntu-jruby.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     steps:
```
eregon: Could you move those 2 jobs to ubuntu.yml now that `ruby/setup-ruby` supports jruby-head?

.github/workflows/ubuntu.yml
```diff
@@ -6,19 +6,18 @@ jobs:
...
+        run: sudo apt install haveged
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install dependencies
+        run: bundle install
+      - name: Run test
+        run: rake
```
eregon: The indentation changes here make it harder to read the diff, could you keep the original indentation?

.github/workflows/windows.yml
MSP-Greg: Re the name, how about 'setup-ruby-win'?  I think that would be clear to most people.

Re RubyInstaller vs ruby-loco mingw, I've set up ruby-loco to test against a master branch.  All test suites are run against install, they're isolated (so all run), and logic is there to determine if any fail.  The mswin build is currently building in a more traditional way (like ruby/ruby CI), but I'm converting it over to match the mingw build.  A lot of repos have been using it for quite a while, including RubyGems.

Off topic, but stepping back, should the Ruby organization support Ruby Windows binaries itself?  Should the 'devkit' be a repo under ruby?  Should ruby/ruby support dll manifests in Windows binaries?  Etc, etc...

.github/workflows/ubuntu-jruby.yml
MSP-Greg: See PR https://github.com/ruby/psych/pull/435.  Tests are failing, so maybe leave them separate for now?

eregon: Right, makes sense.

.github/workflows/ubuntu.yml
MSP-Greg: I could, but it was my impression that proper yaml should have array members indented...

.github/workflows/windows.yml
eregon: `setup-ruby-win` sounds good to me.

I'd be OK to add mingw/mswin builds built by ruby-loco in `ruby/setup-ruby`, if it doesn't add much complexity to the action, PR welcome.

.github/workflows/ubuntu.yml
eregon: Yeah, I'm not sure if there is a recommendation on that, I think both styles were used in `.travis.yml` back then.
I personally prefer the lesser indented variant, so it's only 4-spaces indented for steps (6 starts to feel quite deep already).

eregon:
@hsbt Could you merge this if you agree?
@MSP-Greg There is a conflict in `ubuntu-jruby.yml`

.github/workflows/ubuntu.yml
MSP-Greg: Re indenting, not an issue I feel strongly about.  But, since I changed the indents in `.github/workflows/ubuntu-jruby.yml`, knowing the preference affects what I do with the merge conflict.

My tired old eyes prefer the indented style...

MSP-Greg:
Apologize for not mentioning it previously, but Appveyor is testing on the below.  The 24-x64 & 25-x64 jobs are identical to jobs done on Actions.  The two 32 bit jobs (24 & 25) cannot be done on Actions, as both the Ruby binaries and the gcc tools are unavailable.  I can remove in this PR...

I prefer to remove jobs in AppVeyor & Travis that are run on Actions, as they often backup CI across the whole Ruby account.

```
  matrix:
    - ruby_version: "24"
    - ruby_version: "24-x64"
    - ruby_version: "25"
    - ruby_version: "25-x64"
```

eregon:
I think it's all right to remove redundant jobs in AppVeyor at least.

MSP-Greg:
> I think it's all right to remove redundant jobs in AppVeyor at least.

Removed with additional commit.

eregon:
I'll merge this, I think @hsbt concerns have been addressed.

.github/workflows/windows.yml
eregon: I filed https://github.com/oneclick/rubyinstaller2/issues/169

PR #433, opened 2020-02-13
Remove unnecessary check on version

deivid-rodriguez:
I had a look at the warnings being fixed in #390 to try to come up with a better solution.

Initially, I was able to reproduce the issue by running `bundle` from the root of this repo. The issue happened because my jruby installation had a `psych` installation at: `~/.rbenv/versions/jruby-9.2.9.0/lib/ruby/gems/shared/gems`, and that specific installation _did not_ have the version guard in `versions.rb` that this repository included. So, when running `bundle` first the local gemspec would be loaded which loads the versions.rb file in the repository relatively, then `bundler` would load the "system" version of psych _without_ the version guard, causing the redefinition warnings.

The changes in this PR fix the issue because the local version of `psych` is always loaded.

However, I cannot reproduce on a pristine installation of jruby, because `psych` is not included in the `~/.rbenv/versions/jruby-9.2.9.0/lib/ruby/gems/shared/gems`, so now I'm not sure how to reproduce this anymore... :/

### DIFF
--- a/lib/psych/versions.rb
+++ b/lib/psych/versions.rb
@@ -2,7 +2,7 @@
 # frozen_string_literal: true
 module Psych
   # The version of Psych you are using
-  VERSION = '3.1.0' unless defined?(::Psych::VERSION)
+  VERSION = '3.1.0'
 
   if RUBY_ENGINE == 'jruby'
     DEFAULT_SNAKEYAML_VERSION = '1.23'.freeze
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Just for completeness, the warning can be reproduced by running `bundle install` under jruby on a checkout of this repo _without_ a `Gemfile.lock` file, and while having the same version of `psych` also globally installed.

PR #434, opened 2020-02-13
Use `ruby/setup-ruby` to try fix CI

deivid-rodriguez:


deivid-rodriguez:
I just noticed that this PR is partially duplicated with https://github.com/ruby/psych/pull/432, although it includes less changes and focused specifically on getting CI green.

hsbt:
Thanks!

PR #436, opened 2020-02-26
libyaml 0.2.2 update

MSP-Greg:
1st commit directly copies files from libyaml source, and updates config.h with version info.

Subsequent commits are by file and detail changes needed for Ruby and minimal compile warnings.

Hopefully will make the next update easier...


MSP-Greg:
libyaml 0.2.4 has been released

PR #437, opened 2020-02-26
Fix gem installation

deivid-rodriguez:
Current version is not installable because the gemspec is invalid, since it includes a `.travis.yml` file that no longer exists, so `rake install` fails like this:

```
$ rake install
rake aborted!
WARNING:  See https://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    [".travis.yml"] are not files
```

This PR fixes the issue and adds a small check for it in one of the workflow files.


deivid-rodriguez:
This one should be very easy to review and merge, in case someone is around :)

yuki24:
@deivid-rodriguez Thanks!

deivid-rodriguez:
Thanks @yuki24 :heart:

PR #438, opened 2020-03-03
Fix ArgumentError with leading and trailing underscores in number str…

mthorn:
…ings.

001:0> require 'psych'
true
002:0> '123_'.to_yaml
Traceback (most recent call last):
        ...
        4: from .../lib/ruby/2.7.0/psych/visitors/yaml_tree.rb:279:in `visit_String'
        3: from .../lib/ruby/2.7.0/psych/scalar_scanner.rb:95:in `tokenize'
        2: from .../lib/ruby/2.7.0/psych/scalar_scanner.rb:104:in `parse_int'
        1: from .../lib/ruby/2.7.0/psych/scalar_scanner.rb:104:in `Integer'
ArgumentError (invalid value for Integer(): "123_")
003:0> '0x_123'.to_yaml
Traceback (most recent call last):
        ...
        4: from .../lib/ruby/2.7.0/psych/visitors/yaml_tree.rb:279:in `visit_String'
        3: from .../lib/ruby/2.7.0/psych/scalar_scanner.rb:95:in `tokenize'
        2: from .../lib/ruby/2.7.0/psych/scalar_scanner.rb:104:in `parse_int'
        1: from .../lib/ruby/2.7.0/psych/scalar_scanner.rb:104:in `Integer'
ArgumentError (invalid value for Integer(): "0x_123")

### DIFF
--- a/lib/psych/scalar_scanner.rb
+++ b/lib/psych/scalar_scanner.rb
@@ -101,7 +101,7 @@ module Psych
     ###
     # Parse and return an int from +string+
     def parse_int string
-      Integer(string.gsub(/[,]/, ''))
+      Integer(string.gsub(/[,_]/, ''))
     end
 
     ###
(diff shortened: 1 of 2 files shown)


jasonswett:
I have a question about this one. What's the lowest version of Ruby that this bug is fixed for? This bug seems to be present for me in Ruby 2.7.3 and 2.7.4 but not 3.0.2. Is this bug fix in any 2.7.x or do you have to upgrade to 3.x to get the fix?

casperisfine:
> do you have to upgrade to 3.x to get the fix?

No you can simply add `psych` to your Gemfile, to replace the version shipped with Ruby.

PR #444, opened 2020-03-31
Combine Actions workflows

MSP-Greg:
One file, two jobs - pass and fail


hsbt:
There is no advance about integrate between linux/macOS and Windows for me. I'm okay to merge only linux and macos.

MSP-Greg:
First of all, I'm not combining the files simply because it can be done.

Regardless of the workflow file structure, in the commit and PR pages, jobs are always shown in a listview.

When using multiple workflow files, the Actions page changes its layout.  Sometimes the left pane changes to a treeview, and the main list shows an entry for each workflow file.

I might suggest you compare two tasks, one with the combined workflows, and one with separate:

1. Let's say the vendored libyaml code is changing, and one wants to check all the compile step logs for warnings.

2. Lets assume testing is random, and a few jobs failed.  Let's also assume that it's believed to be a test suite issue.  Find all the failing tests and get the `seed` value, so one can see if the failures repo.

I find inspecting logs to be much easier to do when there is a single workflow file.  Or, a lot fewer mouse clicks to get to step logs.

When would I use separate workflow files:

3. CI has a very a large number of jobs.

4. Jobs run different tasks, e.g. running both Bundler tests and RubyGems tests on a commit/PR.

If you prefer separate workflows, please close.

deivid-rodriguez:
FWIW, this workflow unification look pretty neat to me.

Only thing I would change is the names pass/fail to stable/unstable. Because "fail" can fail but doesn't need to fail. So "unstable" fits the "allowed failure" concept better in my opinion.

MSP-Greg:
Re pass/fail, I think I used that mostly because they're short words, and on some devices the job name is truncated.

I'm fine with pass/fail, stable/unstable, or required/optional

MSP-Greg:
1. I updated the workflow taking into account both the listviews and the Actions page.

2. Ubuntu apt/apt-get is having issues, so I looked for why haveged was installed.  See https://github.com/ruby/psych/commit/0cc4fd116471be4b1f582f6309c4c5ccba7b1555, dated 2016-Nov.  It appears to have been installed to help with JRuby installation, which isn't an issue anymore.  Removed the install.

3. The workflow is using MSP-Greg/setup-ruby-pkgs, mostly just to update the old MinGW gcc on the Windows image.  GitHub is now on their 2nd PR to add an MSYS2 installation, but haven't pushed it yet.  When that is done, the action can be changed to ruby/setup-ruby, as the MSYS2 install will be updated every time they update the image.

JFYI, MSP-Greg/setup-ruby-pkgs exists to help with workflows where cross-platform package installation is needed (like openssl, ragel, etc), and that isn't done here.

hsbt:
I prefer separated workflows that are unix and windows.

deivid-rodriguez:
@hsbt Can you clarify the rejection? All workflows seems to be doing the same thing, so why not merge them all?

hsbt:
In my experience, There is no advance to merge *nix and Windows in CI environment. The issues was caused by PATH environment, Package system and Compiler.

deivid-rodriguez:
I think that used to be the case, but I believe that's been greatly improved recently by @eregon and @MSP-Greg in the official ruby/setup-ruby action. I think this PR passing CI was a good example of that.

eregon:
I agree, I think it's great when the Windows case can be expressed just the same as Linux & macOS, and I think that's becoming a reality for many gems now thanks to fixes in `ruby/setup-ruby` and `setup-ruby-pkgs`.
If the job definition for Windows starts to differ, it's of course always possible to split it to a separate job definition.

In fact, one thing that made me interested in GitHub Actions is it's AFAIK the first free CI including Windows, Linux & macOS, and one can easily define the same job working on all 3 major operating systems.

eregon:
Also I think with this PR the workflow would be significantly easier to maintain, to read and to understand what it does without all that duplication.

eregon:
Similar and related discussion in https://github.com/flori/json/pull/406

MSP-Greg:
Everyone here has been involved in similar discussions elsewhere.  I hate to ruin what's normally a Holiday, but wondering if @tenderlove has an opinion, he's both a ruby/ruby contributor and a maintainer here.  I believe he made the first commit.

In particular, comment https://github.com/ruby/psych/pull/444#issuecomment-610098193

Issue #443, opened 2020-03-25
Quoting strings y/n when dumping?


tisba:
Hey there 👋 

I'm trying to wrap my head around booleans and YAML. We encountered the issue, when we tried to parse a YAML back generated by `YAML.dump()`.

Environment: Ruby 2.7.0p0, 

Example:

```ruby
{"n" => "y"}.to_yaml
```

generates

```yml
---
n: y
```

When we parse this e.g. using https://github.com/go-yaml/yaml it parses as (written as Ruby):

```ruby
{ false => true }
```

When I look at https://yaml.org/type/bool.html, I would assume that the expected generated YAML should be more like this a String rendered:

```yml
---
'n': 'y'
```

Am I missing something?

Some more experiments, in case that's any help, show that `y`, `Y`, `n` and `N` are not quoted as required:

```ruby
# taken from https://yaml.org/type/bool.html
v = %w(y Y yes Yes YES n N no No NO true True TRUE false False FALSE on On ON off Off OFF)

v.each { |x| puts YAML.dump(x) }
# --- y
# --- Y
# --- 'yes'
# --- 'Yes'
# --- 'YES'
# --- n
# --- N
# --- 'no'
# --- 'No'
# --- 'NO'
# --- 'true'
# --- 'True'
# --- 'TRUE'
# --- 'false'
# --- 'False'
# --- 'FALSE'
# --- 'on'
# --- 'On'
# --- 'ON'
# --- 'off'
# --- 'Off'
# --- 'OFF'
```

esotericpig:
According to the [YAML spec](https://yaml.org/spec/1.1/#id862121), keys can be non-strings:

> Since YAML mappings require key uniqueness, representations must include a mechanism for testing the equality of nodes. This is non-trivial since YAML allows various ways to format a given scalar content. For example, the integer eleven can be written as “013” (octal) or “0xB” (hexadecimal). If both forms are used as keys in the same mapping, only a YAML processor which recognizes integer formats would correctly flag the duplicate key as an error.

As you can see here, mapping keys can be integers (and therefore, they can also be booleans, etc.).

If you want a string, you must quote it:

```ruby
Psych.load("---\n1: 2\n")            # => {1=>2}
Psych.load("---\n'1': '2'\n")        # => {"1"=>"2"}
Psych.load("---\ntrue: false\n")     # => {true=>false}
Psych.load("---\n'true': 'false'\n") # => {"true"=>"false"}
```

This is all functioning correctly, per the spec.

[...]

PR #448, opened 2020-04-20
Make 'y' & 'n' into booleans (true/false)

esotericpig:
Issue #443

In order to be compatible with Ruby's original Syck library, Psych made y/n default to strings. However, now, the Syck library is dead, and I think it makes since to make Psych compliant with YAML spec v1.1.

In the issue, this caused a problem when Psych produced a non-compliant file that then broke in a compliant parser in another language. I think it's important that Psych makes YAML files that can be consumed by anything that speaks YAML.

[https://yaml.org/type/bool.html](https://yaml.org/type/bool.html)

However, this is a potentially breaking change. If previous gems relied on y/n as strings, they will now need to be updated to use single/double quotes. This is rare, but should be considered, especially for YAML common as this:

```YAML
MySprite:
  x: 100
  y: 100
  z: 1
```

It will need to be updated to this:

```YAML
MySprite:
  x: 100
  'y': 100
  z: 1
```

Maybe this should be a major version change? However, these gems relied on a broken/non-compliant parser, so it's a difficult decision to make.

Instead of this solution, we could...
- add some type of `strict_compliance` keyword arg option,
- or, output a deprecated warning to users for a couple of versions to update their YAML files.

### DIFF
--- a/lib/psych/scalar_scanner.rb
+++ b/lib/psych/scalar_scanner.rb
@@ -41,9 +41,9 @@ module Psych
           string
         elsif string == '~' || string.match?(/^null$/i)
           nil
-        elsif string.match?(/^(yes|true|on)$/i)
+        elsif string.match?(/^(y|yes|true|on)$/i)
           true
-        elsif string.match?(/^(no|false|off)$/i)
+        elsif string.match?(/^(n|no|false|off)$/i)
           false
         else
           string
--- a/test/psych/test_boolean.rb
+++ b/test/psych/test_boolean.rb
@@ -6,32 +6,44 @@ module Psych
   # Test booleans from YAML spec:
   # http://yaml.org/type/bool.html
   class TestBoolean < TestCase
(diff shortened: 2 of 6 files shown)


esotericpig:
The unsuccessful builds on Windows don't seem to be a problem from my code.

tenderlove:
I think a strict mode would probably be good.  We were talking about it in https://github.com/ruby/psych/pull/426

I'm reluctant to change the behavior without asking people to "opt-in" because people are no doubt depending on this behavior.  Even doing it on a major version bump is a tough sell.  From a consumer's perspective they would think "I have to audit all of my YAML documents when I do this upgrade, but what do I get out of the upgrade?".  Unless we can make the upgrade transparent (IOW users don't need to do anything), I'm not partial to a change like this as the default behavior.

esotericpig:
A strict mode sounds good to me. I'll close this pull request.

PR #457, opened 2020-07-16
Update SNAKEYAML CVE-2017-18640

SzymonKowalczyk:
Update snakeyaml to version 1.26 or higher CVE-2017-18640

### DIFF
--- a/lib/psych/versions.rb
+++ b/lib/psych/versions.rb
@@ -2,7 +2,7 @@
 # frozen_string_literal: true
 module Psych
   # The version of Psych you are using
-  VERSION = '3.1.1'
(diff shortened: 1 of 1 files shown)


lib/psych/versions.rb
```diff
@@ -2,9 +2,9 @@
 # frozen_string_literal: true
 module Psych
   # The version of Psych you are using
-  VERSION = '3.1.0'
+  VERSION = '3.1.1'
```
hsbt, line 5: Can you revert this change?

SzymonKowalczyk: psych version reverted

PR #473, opened 2021-01-29
fix parsing integer values with '_' at the end

OpakAlex:
Hello, We have issue when we parse integer values with `_` at the end, parser removes `_` and then we have just integer. I am pretty sure this is right behaviour for ruby language, but when you are working with customer data, it's important to save it as customer entered.

Some examples:
```
yaml column:  `user_meta`
           values: { "author_id" => "213_312_312_"}, { "project_id" => "1001_"}
```

### DIFF
--- a/lib/psych/scalar_scanner.rb
+++ b/lib/psych/scalar_scanner.rb
@@ -14,10 +14,10 @@ module Psych
               |\.(nan|NaN|NAN)(?# not a number))$/x
 
     # Taken from http://yaml.org/type/int.html
-    INTEGER = /^(?:[-+]?0b[0-1_,]+          (?# base 2)
-                  |[-+]?0[0-7_,]+           (?# base 8)
-                  |[-+]?(?:0|[1-9][0-9_,]*) (?# base 10)
-                  |[-+]?0x[0-9a-fA-F_,]+    (?# base 16))$/x
+    INTEGER = /^(?:[-+]?0b[0-1_,]+               (?# base 2)
+                  |[-+]?0[0-7_,]+                (?# base 8)
+                  |[-+]?(?:\d|[1-9][0-9_,]*[^_]) (?# base 10)
+                  |[-+]?0x[0-9a-fA-F_,]+         (?# base 16))$/x
 
     attr_reader :class_loader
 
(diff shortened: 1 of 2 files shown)


OpakAlex:
@hsbt can you look please?

franzliedke:
@OpakAlex Thanks for fixing this! We encountered this issue on production as well, and I was just about to try my hand at a PR as well. Good thing I checked the list first. :sweat_smile:

lib/psych/scalar_scanner.rb
```diff
@@ -14,10 +14,10 @@ class ScalarScanner
               |\.(nan|NaN|NAN)(?# not a number))$/x
 
     # Taken from http://yaml.org/type/int.html
-    INTEGER = /^(?:[-+]?0b[0-1_,]+          (?# base 2)
-                  |[-+]?0[0-7_,]+           (?# base 8)
-                  |[-+]?(?:0|[1-9][0-9_,]*) (?# base 10)
-                  |[-+]?0x[0-9a-fA-F_,]+    (?# base 16))$/x
+    INTEGER = /^(?:[-+]?0b[0-1_,]+                        (?# base 2)
+                  |[-+]?0[0-7_,]+                         (?# base 8)
+                  |[-+]?(?:0|[1-9]([0-9]|,[0-9]|_[0-9])*) (?# base 10)
```
olleolleolle, line 19: Perhaps this new grouping should also be non-capturing?

```suggestion
                  |[-+]?(?:0|[1-9](?:[0-9]|,[0-9]|_[0-9])*) (?# base 10)
```

Interactive example at: https://rubular.com/r/x8Xufg5TPMDU3X

OpakAlex:
thanks @olleolleolle

lib/psych/scalar_scanner.rb
OpakAlex: @olleolleolle can you check spec?

tenderlove:
Looks like the regex in [the spec](https://yaml.org/type/int.html) technically supports underscore at the end of an integer. But I really think that isn't intentional.  This change seems fine to me.

casperisfine:
I'm afraid this PR introduced a regression: https://github.com/ruby/psych/issues/527

casperisfine:
Nevermind, reading the discussion it seems that it was intentional

PR #476, opened 2021-02-11
Fix custom marshalization with symbolize_names: true

casperisfine:
Reported to me by @peterthesling.

Repo script
```ruby
require 'psych'
class Foo
  def initialize(bar)
    @bar = bar
  end
  def init_with(coder)
    @bar = coder["bar"]
  end
  def encode_with(coder)
    coder["bar"] = @bar
  end
end
p Psych.load(Psych.dump(Foo.new(42))) # => #<Foo:0x00007fea51a63d88 @bar=42>
p Psych.load(Psych.dump(Foo.new(42)), symbolize_names: true) # => #<Foo:0x00007fea51a625a0 @bar=nil>
```

What happens is that the Hash built to be passed to `init_with` get its keys symbolized, which breaks the `init_with`.

cc @tenderlove

### DIFF
--- a/lib/psych/visitors/to_ruby.rb
+++ b/lib/psych/visitors/to_ruby.rb
@@ -339,7 +339,7 @@ module Psych
         list
       end
 
-      def revive_hash hash, o
+      def revive_hash hash, o, tagged= false
         o.children.each_slice(2) { |k,v|
           key = accept(k)
           val = accept(v)
@@ -366,7 +366,7 @@ module Psych
               hash[key] = val
             end
           else
-            if @symbolize_names
+            if !tagged && @symbolize_names
               key = key.to_sym
             elsif !@freeze
               key = deduplicate(key)
@@ -404,7 +404,7 @@ module Psych
 
       def revive klass, node
         s = register(node, klass.allocate)
-        init_with(s, revive_hash({}, node), node)
+        init_with(s, revive_hash({}, node, true), node)
       end
 
       def init_with o, h, node
--- a/test/psych/test_marshalable.rb
+++ b/test/psych/test_marshalable.rb
@@ -51,5 +51,13 @@ module Psych
       assert(PsychCustomMarshalable === loaded)
       assert_equal(2, loaded.foo)
     end
+
+    def test_init_symbolize_names
+      obj = PsychCustomMarshalable.new(1)
+      loaded = Psych.load(Psych.dump(obj), symbolize_names: true)
+
+      assert(PsychCustomMarshalable === loaded)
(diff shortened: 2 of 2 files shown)


casperisfine:
jruby failures seem unrelated.

peterthesling (approved):
Thanks for the fast response!

lib/psych/visitors/to_ruby.rb
```diff
@@ -366,7 +366,7 @@ def revive_hash hash, o
               hash[key] = val
             end
           else
-            if @symbolize_names
+            if !force_string_keys && @symbolize_names
```
whitehat101, line 369: `force_string_keys` is a misleading name. If you have a YAML doc that has Symbols, Numerics, Booleans, etc, as keys then setting `force_string_keys = true` passes those through without modification, it doesn't force them to be strings, it just skips the `#to_sym` call.

whitehat101:
What's the use case for combining custom `init_with` and `encode_with` with  `Psych.load(..., symbolize_names: true)`?

If you have an untyped/untagged document, and you're using YAML as a JSON replacement, `symbolize_names: true` makes sense since it's mimicking the `JSON.load` API, and lets you get symbols without any customization.

But if you do have custom `init_with` and `encode_with`, that's because you want fine control over how your object is (un)marshaled. If your object has an `init_with`, are we expecting  `symbolize_names: true` is ignored? Or are we expecting that `symbolize_names: true` only applies to untagged mappings?

Behavior like this?:
```ruby
doc = <<~YAML
---
baz: !ruby/object:Foo
  bar: 42
YAML

p Psych.load(doc)
# {"baz"=>#<Foo:0x000056141892d218 @bar=42>}
p Psych.load(doc, symbolize_names: true)
# {:baz=>#<Foo:0x0000561418cefe00 @bar=42>}
```

Using `symbolize_names` because you can't or don't want to use symbols in the dumped document?
```ruby
doc = <<~YAML
---
:baz: !ruby/object:Foo
  bar: 42
YAML
p Psych.load(doc)
# {:baz=>#<Foo:0x0000561418cefe00 @bar=42>}

```

lib/psych/visitors/to_ruby.rb
casperisfine: Yes, I wasn't very insprired at that time, I think I'll rename it to `tagged` or something.

casperisfine:
> Or are we expecting that symbolize_names: true only applies to untagged mappings?

Yes, and that was the behavior up to Psych 3.1.0, this PR fixes a regression.

PR #485, opened 2021-05-10
Use test-unit instead of minitest

hsbt:
Psych and RubyGems use `minitest` in https://github.com/ruby/ruby. I hope to migrate them to `test-unit` for the maintainability.


hsbt:
@tenderlove I will merge this. Be careful when writing the new tests with test-unit.

PR #486, opened 2021-05-10
CI - YAML: Avoid 3.0 -> "3" conversion

olleolleolle:
This PR changes the CI configuration slightly, to avoid a float-to-string conversion loss of precision - 3.0 could become "3".


hsbt:
👍  Thanks always!

PR #487, opened 2021-05-10
Use Psych.safe_load by default

tenderlove:
Psych.load is not safe for use with untrusted data.  Too many
applications make the mistake of using `Psych.load` with untrusted data
and that ends up with some kind of security vulnerability.

This commit changes the default `Psych.load` to use `safe_load`.  Users
that want to parse trusted data can use Psych.unsafe_load.

`Psych.load` or `YAML.load` are susceptible to a Remote Code Execution (or RCE) flaw.  The problem isn't with the load functions themselves, but they can be used to abuse other objects that are in the system to escalate to an RCE. [This blog post](https://staaldraad.github.io/post/2019-03-02-universal-rce-ruby-yaml-load/) goes in to details about how YAML can be used in conjunction with RubyGems to execute arbitrary code.  If someone tries to load YAML from an untrusted source, bad actors could use this issue to execute arbitrary code in the target system.

Psych provides a method called "safe_load" which can't be used to launch such an attack.  The proposal here is to change the default such that "load" is now "safe_load".  Of course the drawback is that "safe_load" is much more limited than "load", and making this change could break existing code.

Also it's probably worth mentioning that the YAML load functions could be used with other objects in the system besides RubyGems to escalate to an RCE. Also worth mentioning that anything that can be used to load arbitrary objects (like `Marshal.load`) can be abused in the same way.  So neither changing RubyGems nor changing to a different serialization scheme (like Marshal) would plug this hole.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -234,9 +234,6 @@ require 'psych/class_loader'
 module Psych
   # The version of libyaml Psych is using
   LIBYAML_VERSION = Psych.libyaml_version.join('.').freeze
-  # Deprecation guard
-  NOT_GIVEN = Object.new.freeze
-  private_constant :NOT_GIVEN
 
   ###
   # Load +yaml+ in to a Ruby data structure.  If multiple documents are
@@ -271,12 +268,7 @@ module Psych
   # YAML documents that are supplied via user input.  Instead, please use the
   # load method or the safe_load method.
   #
-  def self.unsafe_load yaml, legacy_filename = NOT_GIVEN, filename: nil, fallback: false, symbolize_names: false, freeze: false
-    if legacy_filename != NOT_GIVEN
-      warn_with_uplevel 'Passing filename with the 2nd argument of Psych.load is deprecated. Use keyword argument like Psych.load(yaml, filename: ...) instead.', uplevel: 1 if $VERBOSE
-      filename = legacy_filename
-    end
-
+  def self.unsafe_load yaml, filename: nil, fallback: false, symbolize_names: false, freeze: false
     result = parse(yaml, filename: filename)
     return fallback unless result
     result.to_ruby(symbolize_names: symbolize_names, freeze: freeze)
@@ -327,27 +319,7 @@ module Psych
   #   Psych.safe_load("---\n foo: bar")                         # => {"foo"=>"bar"}
   #   Psych.safe_load("---\n foo: bar", symbolize_names: true)  # => {:foo=>"bar"}
   #
-  def self.safe_load yaml, legacy_permitted_classes = NOT_GIVEN, legacy_permitted_symbols = NOT_GIVEN, legacy_aliases = NOT_GIVEN, legacy_filename = NOT_GIVEN, permitted_classes: [], permitted_symbols: [], aliases: false, filename: nil, fallback: nil, symbolize_names: false, freeze: false
-    if legacy_permitted_classes != NOT_GIVEN
-      warn_with_uplevel 'Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.', uplevel: 1 if $VERBOSE
-      permitted_classes = legacy_permitted_classes
-    end
-
-    if legacy_permitted_symbols != NOT_GIVEN
-      warn_with_uplevel 'Passing permitted_symbols with the 3rd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_symbols: ...) instead.', uplevel: 1 if $VERBOSE
-      permitted_symbols = legacy_permitted_symbols
-    end
-
-    if legacy_aliases != NOT_GIVEN
-      warn_with_uplevel 'Passing aliases with the 4th argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, aliases: ...) instead.', uplevel: 1 if $VERBOSE
-      aliases = legacy_aliases
-    end
-
-    if legacy_filename != NOT_GIVEN
-      warn_with_uplevel 'Passing filename with the 5th argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, filename: ...) instead.', uplevel: 1 if $VERBOSE
-      filename = legacy_filename
-    end
-
+  def self.safe_load yaml, permitted_classes: [], permitted_symbols: [], aliases: false, filename: nil, fallback: nil, symbolize_names: false, freeze: false
     result = parse(yaml, filename: filename)
     return fallback unless result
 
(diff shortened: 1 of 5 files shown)


tenderlove:
For context, `Psych.safe_load` is more close to JSON loading.  It doesn't support Symbols, Dates, Objects, references, etc.  The only things it supports by default are:

* TrueClass
* FalseClass
* NilClass
* Numeric
* String
* Array
* Hash

stevecrozz:
I support this change, but what do you plan to do with the version number? Feels like a major change.

SamSaffron:
I am extremely supportive of switching to safe by default, even going through a full rename dance here. 

deprecate `safe_load` in favor of `load`
introduce `unsafe_load`

tenderlove:
> lol ... I just read the PR looks like you are very close already, only change I recommend adding is deprecating `safe_load`

I've made `load` slightly different from `safe_load`.  `load` allows all Symbols where `safe_load` doesn't.  I'm not sure if it's a good idea or not.  It's still possible for some symbols to never be garbage collected, but they have to go through a C extension that makes them that way.  However, I suspect many (most?) people use symbols in their YAML and this would trip them up.  It seems like a very rare issue, so maybe allowing symbols is "safe enough"? cc @evanphx

SamSaffron:
>  It seems like a very rare issue, so maybe allowing symbols is "safe enough"?

Yeah sounds like at worst this is allowing some cascade of components to cause memory bloat. Enormously better place to be than RCE friendly by default. 

I guess people like this kind of stuff and tend to use it? 

```
---
:a: 1
```

One concern is that the "load" then is less cross compatible with other frameworks. `:a: 1` in golang/js/rust will be treated as the string `:a`. Maybe `load(str, allow_symbols: true)` and disable by default? Not sure. Would be a far bigger breaking change.

tenderlove:
> One concern is that the "load" then is less cross compatible with other frameworks. :a: 1 in golang/js/rust will be treated as the string :a.

That's true, but it's already the current behavior.  For example:

```ruby
YAML.load(<<-eostr)
---
:a: 1
eostr
```

The above code will result in the same data structure before and after this PR is merged.  Allowing symbols is a compromise for compatibility / non-surprising behavior (as it's what people expect today).  Also, interlanguage compatibility hasn't really been a priority for the library.

I'm leaning towards allowing symbols for `YAML.load` (but *not* `YAML.safe_load`) as I think that will give the best "backwards compatibility plus safety" bang for our buck.

tenderlove:
https://github.com/hashie/hashie/pull/473

k0kubun:
+1 for the interface change.

How about releasing `unsafe_*` variants alone first though? Being able to explicitly rewrite `load` to `safe_load` or `unsafe_load` (where `unsafe_` is really necessary) before switching the behavior of `load` would be useful to minimize the impact of breaking changes by the upgrade.

lib/psych.rb
```diff
@@ -262,16 +262,16 @@ module Psych
...
   # Raises a TypeError when `yaml` parameter is NilClass
   #
   # NOTE: This method *should not* be used to parse untrusted documents, such as
   # YAML documents that are supplied via user input.  Instead, please use the
-  # safe_load method.
+  # load method or the safe_load method.
   #
-  def self.load yaml, legacy_filename = NOT_GIVEN, filename: nil, fallback: false, symbolize_names: false, freeze: false
+  def self.unsafe_load yaml, legacy_filename = NOT_GIVEN, filename: nil, fallback: false, symbolize_names: false, freeze: false
     if legacy_filename != NOT_GIVEN
```
k0kubun, line 275: and I think you don't need to support the legacy argument in the new method.

tenderlove:
> How about releasing `unsafe_*` variants alone first though? Being able to explicitly rewrite `load` to `safe_load` or `unsafe_load` (where `unsafe_` is really necessary) before switching the behavior of `load` would be useful to minimize the impact of breaking changes by the upgrade.

Yes, that is a great idea. I will do it!

tenderlove:
I shipped this in v4.0.0.  Thanks!

lzap:
Friday, our CI is red. Only if there was a better week day to push big releases out, hmmm 🤔 

Well, a mistake in our `Gemfile`, we will fix it, and the code. Thank you for taking care of safety of all of us! Cheese, I mean, cheers! :-)

eregon:
Seeing the fallout of this change (notably at https://bugs.ruby-lang.org/issues/17866, particularly aliases) I'm not sure it's worth it.
If people deal with user input, they should know they can never blindly use `X.load` but should validate the input first, or if the deserializer is safe then potentially validating after.
Also it seems rather uncommon to have user-supplied YAML (e.g. in a web app), isn't it?
AFAIK the most common case for YAML is config, and there it should be the choice of whatever reads that YAML to decide what to allow.

One concern is so far X.dump and X.load in Ruby were able to round trip data and this change breaks that.

Also it seems weird that `JSON.load` is kept as the roundtrip one (and `JSON.parse` is the safe variant) but a completely different approach (no common names and meaning) is used for YAML. Maybe the issue is that `YAML.parse` already exists?

lzap:
I do not understand one thing tho, I'd expect that Ruby symbols are allowed even for unsafe loading, but it looks like after we updated to psych 4.0 this is not the case. I am still seeing `#<Psych::DisallowedClass: Tried to load unspecified class: Symbol>`. We do have a bunch of YAML configuration files with only Ruby symbols and load them this way:

    YAML.load_file(config_file)

The change we need to do was:

    YAML.load_file(config_file, permitted_classes: [Symbol])

Was this intentional behavior?

Unfortunately, there are exceptions across our dependencies (e.g. rails_i18n) so pinning to 3.3 is the only option for now until all projects are fixed.

lib/psych.rb
```diff
@@ -363,6 +335,46 @@ def self.safe_load yaml, legacy_permitted_classes = NOT_GIVEN, legacy_permitted_
...
+  # When the optional +symbolize_names+ keyword argument is set to a
+  # true value, returns symbols for keys in Hash objects (default: strings).
+  #
+  #   Psych.load("---\n foo: bar")                         # => {"foo"=>"bar"}
+  #   Psych.load("---\n foo: bar", symbolize_names: true)  # => {:foo=>"bar"}
+  #
+  # Raises a TypeError when `yaml` parameter is NilClass.  This method is
+  # similar to `safe_load` except that `Symbol` objects are allowed by default.
+  #
+  def self.load yaml, permitted_classes: [Symbol], permitted_symbols: [], aliases: false, filename: nil, fallback: nil, symbolize_names: false, freeze: false
```
simi, line 368: :thinking: I am just wondering, is there any particular security reason to not enable aliases by default?

simi: @tenderlove any idea (I'm tagging you as an author, I hope no harm done :pray:)?

tenderlove: Aliases allow you to construct self referencing data structures which can lead to a DoS attack (think a recursive function that never ends)

Sarkar44 (commented):
approve

PR #488, opened 2021-05-13
Introduce `Psych.unsafe_load`

tenderlove:
In future versions of Psych, the `load` method will be mostly the same
as the `safe_load` method.  In other words, the `load` method won't
allow arbitrary object deserialization (which can be used to escalate to
an RCE).  People that need to load *trusted* documents can use the
`unsafe_load` method.

This commit introduces the `unsafe_load` method so that people can
incrementally upgrade.  For example, if they try to upgrade to 4.0.0 and
something breaks, they can downgrade, audit callsites, change to
`safe_load` or `unsafe_load` as required, and then upgrade to 4.0.0
smoothly.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -271,7 +271,7 @@ module Psych
   # YAML documents that are supplied via user input.  Instead, please use the
   # safe_load method.
   #
-  def self.load yaml, legacy_filename = NOT_GIVEN, filename: nil, fallback: false, symbolize_names: false, freeze: false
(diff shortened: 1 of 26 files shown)


DannyBen:
This change causes a major ripple effect. I am sure it had its reasons, but `safe_load` already existed for people who needed more security. Now people who need to support both Ruby 3.1/Psych 4 as well as older versions, must use some Ruby version checking or `rescue ArgumentError`.

PR #502, opened 2021-05-29
Adding flag :noImplicitTags to dump options

Mai-Lapyst:
Adding a :noImplicitTags flag to dump options to prevent psych from outputing most of the generated tags (mostly `!ruby`).
Also adds a few more lookups into `Psych.dumb_tags` for certain builtin types to allow users to specify their own tags for them.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -489,6 +489,10 @@ module Psych
   #
   #                           Default: <tt>false</tt>.
   #
+  # [<tt>:no_implicit_tags]   Prevents Psych from adding implicit tags to objects.
+  #
+  #                           Default: <tt>false</tt>
+  #
   # Example:
   #
   #   # Dump an array, get back a YAML string
@@ -561,6 +565,10 @@ module Psych
   # [<tt>:header</tt>]        Write <tt>%YAML [version]</tt> at the beginning of document.
   #
   #                           Default: <tt>false</tt>.
+  #
+  # [<tt>:no_implicit_tags]   Prevents Psych from adding implicit tags to objects.
+  #
+  #                           Default: <tt>false</tt>
   #
   # Example:
   #
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -152,7 +152,7 @@ module Psych
 
       def visit_Object o
         tag = Psych.dump_tags[o.class]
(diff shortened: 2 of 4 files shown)


olleolleolle:
Minor: A  general note, the other `@options` are snake_case (e.g. line_width).

Mai-Lapyst:
@olleolleolle Didn't know that, but thanks for the feedback; will fix it ^^

olleolleolle:
I am guessing a unit test or two for this option would help the Pull Request.

tenderlove:
Why do you need this?  It breaks round tripping (IOW data structures emitted with Psych will not equal the same data structures that are loaded).  If you have unwanted objects in your data structure, it seems better to replace them with what you want?  I just don't understand the use case.

Mai-Lapyst:
The use case is, when using psych to generate yaml that is then loaded into another language.
Some parsers out there fail when they find unknown tags, or parse the data as an hashmap instead
of the correct internal representation. For example, the datetime object is annotated with
`!ruby/object:DateTime` but when the reader expect for example `!datetime` then the reciever dosnt
parse the data right.

Another usecase are document specifications like RAML that may disallow the use of tags they dont define,
thus it would classify yaml output from psych as "malformed". (Which is the case for RAML)

Also it's always nicer to have an option for these kind of things so people can use psych like they want
and dont need to use an extra script that converts the psych output to that what they wanted in the first place.

And I dont think it will break round tripping since this is an option that isn't enabled by default, so
existing applications dont break. Also new applications don't have an round tripping propblem as long as they dont touch
the `:no_implicit_tags` option or `Psych.dump_tags`.

tenderlove:
Sorry, I'm not convinced.  "Date Times" have different representations in different languages.  Psych can't know what language you'll be parsing the output with.  Since you are using Ruby to output the data, it knows how to "round trip" the data.

For example `Psych.load(Psych.dump(DateTime.now))` will return the same date time object, but `Psych.load(Psych.dump(DateTime.now, noImplicitTags: true))` cannot return the same object.  If you want to dump something custom for a different language, then convert the object to something else before giving it to Psych.

For example, rather than `Psych.dump(DateTime.today)`, do

```ruby
x = DateTime.now
Psych.dump(x.strftime("%Y-%m-%d"))
```

I realize this will dump a string, but it's outside the scope of this project to predict the format you need on the reading side.

Issue #371, opened 2018-08-21
Psych.dump convert emoji characters into unicode


neumayr:
Hey!  🙌 

The origin of this issue is [`i18n-tasks`](https://github.com/glebm/i18n-tasks) which [delegates](https://github.com/glebm/i18n-tasks/issues/291#issuecomment-414433435) YAML dumping to this library.

> I've spotted that `i18n-tasks normalize` converts emoji characters into unicode which makes them much harder to read and change afterward. IMHO, this should be necessary because `yaml` is capable to understand plain emoji chars. What do you think about this one?

##### Before
```yml
lang:
  de-de: 🇩🇪 German
  en-uk: 🇺🇸 English
  header: 🌐 Select Language
```

##### After `i18n-tasks normalize` / `Psych.dump` 
```
lang:
  de-de: "\U0001F1E9\U0001F1EA German"
  en-uk: "\U0001F1FA\U0001F1F8 English"
  header: "\U0001F310 Select Language"
```

#### Converter tool
* https://r12a.github.io/app-conversion/

✨ Please let emojis live in `yaml` files. 💙 💚 💛 💜

---

**Note** This issue was moved from https://github.com/glebm/i18n-tasks/issues/291

jkostolansky:
Hi, has there been any progress made on this?

forelabs:
Upvote for having emojis in yaml files. It's 2020, emojis are basically everywhere. Would be nice to have this supported.

DTrejo:
FYI, still happens with some emoji like 🔧. Thanks for the Fix @bibendi !

PR #508, opened 2021-06-09
Implement emoji support

Liberatys:
This pull request addresses an issue with 4 byte unicode characters. It uses code from a pr which was filed against **tarantool** in response to https://github.com/tarantool/tarantool/issues/4090.

I'm well aware that this probably should belong to the **libyaml** repository. I wanted to open it here first to have a base to work from. Response time in libyaml seems to be somewhat long and thus I think this is a feature we should implement asp. This could unlock faster automated translations in Rails(as I18n hands over yaml generation to this gem).

### DIFF
--- a/ext/psych/extconf.rb
+++ b/ext/psych/extconf.rb
@@ -7,34 +7,74 @@ require 'fileutils'
 
 dir_config 'libyaml'
 
-if enable_config("bundled-libyaml", false) || !(find_header('yaml.h') && find_library('yaml', 'yaml_get_version'))
-  # Embed libyaml since we could not find it.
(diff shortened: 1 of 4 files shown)


coorasse:
Fixes #371 and  https://github.com/glebm/i18n-tasks/issues/291 ?

Liberatys:
It should! We can add some more tests to ensure that it works for the most commonly used Unicode characters

tenderlove:
Seems to break CI 😅.  I really like the idea of supporting emoji, but I'm somewhat worried about supporting effectively a fork of libyaml.  But I really do like the idea of supporting emoji 🤔

PR #515, opened 2021-08-04
Add quotes to the strings "y" and "n"

### LINKED ISSUE #443: Quoting strings y/n when dumping?

tenderlove:
'y' and 'n' are kind of ambiguous.  Syck treated y and n literals in
YAML documents as strings.  But this is not what the YAML 1.1 spec says.
YAML 1.1 says they should be treated as booleans.  When we're dumping
documents, we know it's a string, so adding quotes will eliminate the
"ambiguity" in the emitted document

Fixes #443

### DIFF
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -272,6 +272,8 @@ module Psych
           tag   = 'tag:yaml.org,2002:str'
           plain = false
           quote = false
+        elsif o == 'y' || o == 'n'
+          style = Nodes::Scalar::DOUBLE_QUOTED
         elsif @line_width && o.length > @line_width
           style = Nodes::Scalar::FOLDED
         elsif o =~ /^[^[:word:]][^"]*$/
--- a/test/psych/test_string.rb
+++ b/test/psych/test_string.rb
@@ -17,6 +17,19 @@ module Psych
       end
     end
 
+    # 'y' and 'n' are kind of ambiguous.  Syck treated y and n literals in
+    # YAML documents as strings.  But this is not what the YAML 1.1 spec says.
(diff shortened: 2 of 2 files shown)


simi:
What about this case? Would it be possible to quote it as well?

```ruby
YAML.dump('A' => '2020-12-31')
=> "---\nA: 2020-12-31\n"

YAML.load(YAML.dump('A' => '2020-12-31'))
=> Tried to load unspecified class: Date (Psych::DisallowedClass)
```

Issue #520, opened 2021-09-20
unsupported Java version "16", defaulting to 1.7 [ERROR] Internal error...


AndyObtiva:
- MacOS Big Sur 11.2.3 x64
- Processor 2.5 GHz Quad-Core Intel Core i7 x64
- rvm 1.29.12-next
- java version "16.0.2" 2021-07-20 Java(TM) SE Runtime Environment (build 16.0.2+7-67) Java HotSpot(TM) 64-Bit Server VM (build 16.0.2+7-67, mixed mode, sharing)
- jruby 9.2.19.0 (2.5.8) 2021-06-15 55810c552b Java HotSpot(TM) 64-Bit Server VM 16.0.2+7-67 on 16.0.2+7-67 +jit [darwin-x86_64]
- gem version 3.2.27 (after running `gem update --system`)

```
andymaleh@Andys-MacBook-Pro snowboard_utah % gem install psych -v4.0.1 
Fetching psych-4.0.1-java.gem
  jar dependencies for psych-4.0.1-java.gemspec . . .
unsupported Java version "16", defaulting to 1.7
[ERROR] Internal error: org.jruby.exceptions.RaiseException: (LoadError) library `java' could not be loaded: java.lang.reflect.InaccessibleObjectException: Unable to make protected native java.lang.Object java.lang.Object.clone() throws java.lang.CloneNotSupportedException accessible: module java.base does not "opens java.lang" to unnamed module @4d96250 -> [Help 1]
org.apache.maven.InternalErrorException: Internal error: org.jruby.exceptions.RaiseException: (LoadError) library `java' could not be loaded: java.lang.reflect.InaccessibleObjectException: Unable to make protected native java.lang.Object java.lang.Object.clone() throws java.lang.CloneNotSupportedException accessible: module java.base does not "opens java.lang" to unnamed module @4d96250
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:121)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:863)
	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:288)
	at org.apache.maven.cli.MavenCli.main(MavenCli.java:199)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:78)
[...]

Issue #479, opened 2021-02-24
JRuby tests failing


headius:
For about the past month the JRuby runs in CI have been failing:

https://github.com/ruby/psych/actions/workflows/ubuntu-jruby.yml

I have not looked into any of the failures but I suspect a couple things:

* No guard against running ractor tests, so they fail due to no ractor support
* Recent fixes to the C ext have not been translated to the JRuby ext

JRuby's branches for 9.2 and master are currently failing because we updated the psych gem and now its tests fail.

headius:
I see now the most recent CI runs for JRuby don't even manage to run the tests. Something fails along the way in Bundler.

Here is a JRuby build showing the failures introduced recently:

https://travis-ci.com/github/jruby/jruby/jobs/486061152

headius:
Note the ractor tests do not show up in the link above, but when I update JRuby's local copy of the psych tests (which includes the new ractor tests) they do attempt to run and fail with missing method errors.

olleolleolle:
@headius 👋 Just a quick check, https://github.com/ruby/psych/actions/workflows/ubuntu-jruby.yml looks green, am I misinterpreting the hue of the indicators?

hsbt:
The failing tests are skipped by #524.

We will remove them when resolving with JRuby.

PR #524, opened 2021-10-12
Only skip specific failures on JRuby instead of ignoring the whole CI feedback

deivid-rodriguez:
This PR is just some experiment to proof that https://github.com/jruby/ruby-maven/pull/2 fixes:

* Issue reported at https://github.com/ruby/psych/issues/479#issuecomment-785469484, where jruby tests fail to even get started in CI (`bundle install` fails).
* https://github.com/ruby/psych/issues/520.
* https://github.com/jruby/ruby-maven/issues/3.
* https://github.com/rubygems/rubygems/issues/4685.

### DIFF
--- a/.github/workflows/ubuntu-jruby.yml
+++ b/.github/workflows/ubuntu-jruby.yml
@@ -16,9 +16,9 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
       - name: Install dependencies
-        run: gem install ruby-maven power_assert test-unit rake-compiler 
+        run: bundle install --jobs 1
       - name: compile
-        run: rake compile
+        run: bundle exec rake compile
       - name: test
         continue-on-error: true
-        run: rake test
+        run: bundle exec rake test
--- a/.github/workflows/ubuntu-jruby.yml
+++ b/.github/workflows/ubuntu-jruby.yml
@@ -20,5 +20,4 @@ jobs:
       - name: compile
         run: bundle exec rake compile
       - name: test
-        continue-on-error: true
         run: bundle exec rake test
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
Alright so dependent PRs are moving forward, and this is now only missing a `ruby-maven` release.

However, the gem became uninstallable again on `jruby` after https://github.com/ruby/psych/pull/525, since gemified `stringio` does not yet support jruby (WIP for that is https://github.com/ruby/stringio/pull/21).

headius:
This would be worth revisiting with latest SnakeYAML. On master, I now see 5F7E. Some of these may be valid differences (SnakeYAML appears to report line/column positioning a bit differently) but several appear to be fixable errors or bugs. We also addressed the stringio gemification for JRuby 9.4.

deivid-rodriguez:
@headius I completely lost context of this PR. I think the original issues I wanted to address are now fixed, but a few others remain. I changed the scope of this PR to:

* Use Bundler to install dependencies and run tests.
* Only skip the particular tests that are failing instead of ignoring the whole test run outcome. I think this should help preventing new regressions.

Let's see what CI says.

deivid-rodriguez:
Except for the specs that are marked as pending, everything is now green on JRuby, so I set this as ready for review.

headius:
@deivid-rodriguez Very nice! Good idea to skip some tests that are problematic... some might not be fixable due to minor differences in SnakeYAML (positioning information etc).

@tenderlove Any objections to merging this PR? It marks a number of tests as pending, but I'm not sure they are fixable and it would be good to have a green-ish run so we can see when things regress.

hsbt:
@deivid-rodriguez 👍 Thanks!

PR #534, opened 2022-01-13
Don't require `strscan` unnecessarily

deivid-rodriguez:
It does not seem needed, and it's causing issues on Windows when uninstalling `strscan`, because strscan's shared library being used when RubyGems tries to remove it (because its loaded through Psych, which RubyGems uses for loading configuration).

Partially fixes https://github.com/rubygems/rubygems/issues/5274.

### DIFF
--- a/lib/psych/scalar_scanner.rb
+++ b/lib/psych/scalar_scanner.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require 'strscan'
 
 module Psych
   ###


hsbt:
👍

darix:
can we get this backported into 3.1?

deivid-rodriguez:
I think it's not even released, so requesting a backport seems premature.

deivid-rodriguez:
By the way, any plans for a release @hsbt? 😃

deivid-rodriguez:
It's out now, thanks so much @hsbt ❤️

darix:
do we need an extra bug on bugs.ruby-lang.org for the backport or is asking here enough? :)

hsbt:
I will add psych-4.0.4 to https://github.com/ruby/ruby/pull/5608

Issue #273, opened 2016-02-25
Ruby 2.1/2.2/2.3 yaml parsing anomaly for comma separated integers


guidoiaquinti:
A string of digits separated by comma (without quotes) is evaluated as integer. Is it intentional?

```
irb(main):001:0> Psych.load("key: 123,456")
=> {"key"=>123456}

irb(main):002:0> Psych.load("key: 123456,7890")
=> {"key"=>1234567890}
```

I understand that the first example can be seen as a number in American notation but the 2nd example is not following any notation/standard that I'm aware of. I was expecting to have both of the values evaluated as string. 

With the quotes it works as expected:

```
irb(main):003:0> Psych.load("key: \"123,456\"")
=> {"key"=>"123,456"}

irb(main):004:0> Psych.load("key: \"123456,7890\"")
=> {"key"=>"123456,7890"}
```

Thanks

tenderlove:
Apparently [Japan and China will separate at four digits](http://www.statisticalconsultants.co.nz/blog/how-the-world-separates-its-digits.html), but to be honest Psych internals don't care, they just remove all commas and cast to a number.  The only reason for this is legacy support for Syck documents.  I agree that both values should evaluate to strings, but I'm not sure how to do that and support old formats.

Jell:
I've hit this issue today after trying to parse some YAML output by another tool (kustomize build in my case, but not that relevant). I think as YAML becomes more and more ubiquitous it might be a good idea to provide an option for "strict" parsing according to the standard?

It seems to me that we use those regexp to decide whether or not to parse a Scalar as a Float or Integer:
https://github.com/ruby/psych/blob/9f8c3655376aecec6056adb2bb12665971980f9f/lib/psych/scalar_scanner.rb#L11-L20

I did this patch on my side as a workaround, which solved the issue I had:

```ruby
Psych::ScalarScanner.send(:remove_const, "INTEGER")
Psych::ScalarScanner::INTEGER = /^(?:[-+]?0b[0-1_]+          (?# base 2)
                                    |[-+]?0[0-7_]+           (?# base 8)
                                    |[-+]?(?:0|[1-9][0-9_]*) (?# base 10)
                                    |[-+]?0x[0-9a-fA-F_]+    (?# base 16))$/x
```

(basically replacing the regexp with a subset of the standard one from from http://yaml.org/type/int.html)

Would it be possible to have say two constants:

```ruby
INTEGER_ENRICHED=/.../
INTEGER_STANDARD=/.../
[...]

PR #537, opened 2022-01-14
Add strict_integer option to parse numbers with commas as strings

sethboyles:
Addresses #273

This PR introduces a `strict_integer` option to the load/parsing methods, allowing for closer adherence to the YML spec, which specifies that numbers including commas should be parsed as strings.  More details in the above linked issue.


tenderlove (approved):
LGTM. I'll merge when the build comes back green.  Thanks for doing this!

sethboyles:
Thanks for the merge!  Any idea when we will see this released?

Issue #535, opened 2022-01-13
Should we unbundle libyaml?


tenderlove:
We talked about unbundling libraries at the dev meeting, and I couldn't remember why we bundled libyaml.

I found the ticket [here](https://bugs.ruby-lang.org/issues/7375).  I guess the problem is that RubyGems depends on Psych, and Psych depends on libyaml, so we need to make sure you can install Psych so that you can install RubyGems.  Essentially you can't use RubyGems unless libyaml is installed.

If Ruby depends on RubyGems, and RubyGems depends on Psych, then Ruby depends on libyaml.

I don't like bundling libraries, but I don't think we should make a hard dependency on libyaml being installed.  In other words, I think we should continue to embed libyaml unless there is a different solution.

Any opinions? @hsbt @unak

hsbt:
Is it related with libffi issue?

Basically, I agree with unbundle it. But We should confirm to build psych with libyaml package provided by vcpkg in mswin environment.

boof:
Just a heads up because we stumbled upon this while having the libyaml-dev package install on Ubuntu LTS. When unbundled this will cause the whitespace issue (https://github.com/yaml/libyaml/pull/186) on Ubuntu LTS again since it still uses an old version w/o the fix.

PR #541, opened 2022-02-04
Remove bundled libyaml

hsbt:
Fixes https://github.com/ruby/psych/issues/535


.github/workflows/windows.yml
```diff
@@ -11,12 +11,18 @@ jobs:
...
         with:
           ruby-version: ${{ matrix.ruby }}
           mingw: _upgrade_
       - name: Install dependencies
         run: bundle install
+      - name: Compile
+        run: rake compile -- --with-libyaml-dir=C:/vcpkg/installed/x64-windows
       - name: Run test
         run: rake
+k
```
olleolleolle, line 28:
```suggestion
```

Typo?

hsbt: oops.

colszowka:
I only now realized there is already a PR for this on top of the original #535 - I would like to suggest not to merge this because there is a bug in the libyaml packaged in the currently used Ubuntu/Debian versions, and if I understand correctly merging this would render the fix (of not installing libyaml-dev) no longer working, see also https://github.com/rbenv/ruby-build/issues/1950 (correct me if I'm wrong, maybe I'm also misunderstanding something about this or how psych and Ruby work together here)

.github/workflows/windows.yml
```diff
@@ -11,12 +11,22 @@ jobs:
         ruby: [ mingw, mswin, 3.0, 2.7, 2.6, 2.5, 2.4 ]
     steps:
       - uses: actions/checkout@v2
+      - name: Install libraries with vcpkg
+        run: |
+          vcpkg --triplet x64-windows install libyaml
+      - name: link libraries
+        run: |
+          for %%I in (C:\vcpkg\installed\x64-windows\bin\*.dll) do (
+            if not %%~nI == readline mklink %%~nxI %%I
```
nobu, line 20:
```suggestion
           mklink %%~nxI %%I
```
Not installing readline.

.github/workflows/windows.yml
```diff
@@ -11,12 +11,26 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - name: Install libraries with vcpkg
+        run: |
+          vcpkg --triplet x64-windows install libyaml
+      - name: link libraries
+        run: |
+          for %%I in (C:\vcpkg\installed\x64-windows\bin\*.dll) do (
+            mklink %%~nxI %%I
+          )
```
nobu, line 21: Need to install libyaml with vcpkg only on mswin, I think.
```suggestion
      - name: Install libraries with vcpkg
        run: |
          vcpkg --triplet x64-windows install libyaml
        if: ${{ matrix.ruby != 'mswin' }}
      - name: link libraries
        run: |
          for %%I in (C:\vcpkg\installed\x64-windows\bin\*.dll) do (
            mklink %%~nxI %%I
          )
        if: ${{ matrix.ruby != 'mswin' }}
```

Issue #552, opened 2022-03-30
Ruby can't compile with psych when libyaml isn't installed


eileencodes:
Hello! Since I pulled Ruby master yesterday I'm unable to configure or run make successfully due to an error that the digest file is missing. This only happens when libyaml isn't installed - if I install libyaml then everything works correctly. However the code that's failing is supposed to download libyaml and instead raises the following error:

```
/Users/eileencodes/open_source/ruby/tool/extlibs.rb:6:in `require': cannot load such file -- digest (LoadError)
	from /Users/eileencodes/open_source/ruby/tool/extlibs.rb:6:in `<top (required)>'
	from /Users/eileencodes/open_source/ruby/ext/psych/extconf.rb:22:in `require_relative'
	from /Users/eileencodes/open_source/ruby/ext/psych/extconf.rb:22:in `<top (required)>'
	from ./ext/extmk.rb:214:in `load'
	from ./ext/extmk.rb:214:in `block in extmake'
	from /Users/eileencodes/open_source/ruby/lib/mkmf.rb:324:in `open'
	from ./ext/extmk.rb:210:in `extmake'
	from ./ext/extmk.rb:572:in `block in <main>'
	from ./ext/extmk.rb:568:in `each'
	from ./ext/extmk.rb:568:in `<main>'
make[1]: *** [ext/psych/exts.mk] Error 1
make[1]: *** Waiting for unfinished jobs....
```

I know that there were a lot of recent changes to psych and libyaml is no longer bundled with it. On my machine where libyaml is not installed when I run `make` on my Ruby checkout (or `configure`) it will hit this line https://github.com/ruby/ruby/blob/master/ext/psych/extconf.rb#L22 which loads `tool/extlibs.rb`. That file then has a `require 'digest'` which is where it is failing to build and cannot load the file. Since it fails to require the digest file, the `download_cache` is never used. I can consistently reproduce if I don't have libyaml installed. If I manually install it with homebrew, Ruby builds fine. 

Machine details:

1) MacOS Monterey, M1 Max
2) Ruby master as of github.com/ruby/ruby/0cb43034b416aaf28bf57ee34bc095fdcb659f33

Steps to reproduce:

1) Uninstall libyaml or ensure it is not installed
2) `make clean`
3) Run configure script, mine below:
```
autoconf
[...]

PR #562, opened 2022-05-26
Fix libyaml download failure rescue under miniruby

XrXr:
I tried to build Ruby on a system without libyaml today and realized
that my attempt from https://github.com/ruby/psych/pull/557 doesn't
fix the error in https://github.com/ruby/psych/issues/552. I still got
the same LoadError from digest.

Since `LoadError` is not a StandardError, a plain `rescue` doesn't catch
it. Catch `LoadError` explicitly instead and reduce the scope of the
`begin` block.

I tested this change in a Ruby build on macOS without libyaml installed
and confirmed that `make` continues with a warning instead of aborting:

    *** Following extensions are not compiled:
    psych:
            Could not be configured. It will not be installed.
            ...

### DIFF
--- a/ext/psych/extconf.rb
+++ b/ext/psych/extconf.rb
@@ -19,15 +19,17 @@ if yaml_source == true
   # search the latest libyaml source under $srcdir
   yaml_source = Dir.glob("#{$srcdir}/yaml{,-*}/").max_by {|n| File.basename(n).scan(/\d+/).map(&:to_i)}
   unless yaml_source
-    download_failure = "failed to download libyaml source"
+    download_failure = "failed to download libyaml source. Try manually installing libyaml?"
     begin
       require_relative '../../tool/extlibs.rb'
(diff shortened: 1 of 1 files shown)


XrXr:
For only graceful failures on the Ruby side, we could apply:

```diff
diff --git a/ext/extmk.rb b/ext/extmk.rb
index a440af27fc..fd6518fa6e 100755
--- a/ext/extmk.rb
+++ b/ext/extmk.rb
@@ -224,7 +224,7 @@ def extmake(target, basedir = 'ext', maybestatic = true)
        end
       rescue SystemExit
        # ignore
-      rescue => error
+      rescue StandardError, ScriptError => error
         ok = false
       ensure
        rm_f "conftest*"
```

This PR gives better error message in addition to graceful failures. We may want to change extmk.rb for similar issues in the future, though.

XrXr:
I'm going to merge this since it should fix https://bugs.ruby-lang.org/issues/18790 for people that don't have libyaml installed on their system. Feel free to revert/ make changes if you notice something off.

PR #567, opened 2022-07-21
Raise specific error when aliases are not enabled

### LINKED ISSUE #565: Parsing aliases `aliases: false` should not raise `Psych::BadAlias`
Simple minimal example:

```ruby
#!/usr/bin/ruby

require "yaml"

yaml_with_aliases = <<~YAML
---
a: &ABC
  k1: v1
  k2: v2
b: *ABC
YAML

p YAML.load(yaml_with_aliases)
```

This raises an error: `Unknown alias: ABC (Psych::BadAlias)`

Raising an error in this case sense, but I think `Psych::BadAlias` is an incorrect error type. It communicates an incorrect impression that the issue is with the structure of the YAML, not the configuration of the `YAML`/`Psych`. The message makes it even worse.

Perhaps there should be a more specific error type, like `Psych::AliasParsingDisabled`. The message could then mention that if you want anchors/aliases to work, that you should passes `aliases: true`.

Thoughts?

amomchilov:
Fixes #565

Introduces a new error, `Psych::AliasesNotEnabled`. For backwards compatibility, it's a subclass of `Psych::BadAlias`, so that old `rescue` blocks keeping working as intended.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -307,7 +307,7 @@ module Psych
   # A Psych::DisallowedClass exception will be raised if the yaml contains a
   # class that isn't in the +permitted_classes+ list.
   #
-  # A Psych::BadAlias exception will be raised if the yaml contains aliases
+  # A Psych::AliasesNotEnabled exception will be raised if the yaml contains aliases
   # but the +aliases+ keyword argument is set to false.
   #
   # +filename+ will be used in the exception message if any exception is raised
--- a/test/psych/test_merge_keys.rb
+++ b/test/psych/test_merge_keys.rb
@@ -117,7 +117,7 @@ development:
 bar:
   << : *foo
       eoyml
-      exp = assert_raise(Psych::BadAlias) { Psych.load yaml }
+      exp = assert_raise(Psych::BadAlias) { Psych.load(yaml, aliases: true) }
       assert_match 'foo', exp.message
     end
 
--- a/test/psych/test_safe_load.rb
+++ b/test/psych/test_safe_load.rb
@@ -19,18 +19,31 @@ module Psych
       end
     end
 
-    def test_no_recursion
-      x = []
-      x << x
+    def test_raises_when_alias_found_if_alias_parsing_not_enabled
+      yaml_with_aliases = <<~YAML
+        ---
+        a: &ABC
+          k1: v1
+          k2: v2
+        b: *ABC
+      YAML
+
       assert_raise(Psych::BadAlias) do
-        Psych.safe_load Psych.dump(x)
+        Psych.safe_load(yaml_with_aliases)
       end
     end
 
-    def test_explicit_recursion
-      x = []
-      x << x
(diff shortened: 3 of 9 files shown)


lib/psych.rb
```diff
@@ -307,7 +307,7 @@ def self.unsafe_load yaml, filename: nil, fallback: false, symbolize_names: fals
   # A Psych::DisallowedClass exception will be raised if the yaml contains a
   # class that isn't in the +permitted_classes+ list.
   #
-  # A Psych::BadAlias exception will be raised if the yaml contains aliases
+  # A Psych::AliasesNotEnabled exception will be raised if the yaml contains aliases
   # but the +aliases+ keyword argument is set to false.
```
amomchilov, line 311: This is a breaking change. Should it stay like this and wait until a major version bump, or should I make `AliasesNotEnabled` a subclass of `BadAlias` (to make the change non-breaking)?

test/psych/test_safe_load.rb
```diff
@@ -22,7 +22,7 @@ class Foo; end
     def test_no_recursion
       x = []
       x << x
-      assert_raise(Psych::BadAlias) do
+      assert_raise(Psych::AliasesNotEnabled) do
         Psych.safe_load Psych.dump(x)
       end
     end
```
amomchilov, line 28: We should fix this test, because I don't think it really makes sense:

Any alias will make this raise `Psych::AliasesNotEnabled`, because `alias: true` wasn't specified. The recursion is a red herring, any alias will trigger this exception to be thrown (and thus the assertion to pass).

lib/psych.rb
tenderlove: You should make it a subclass.  I think `BadAlias` could still be a problem in the case that there is an alias with no corresponding anchor

test/psych/test_safe_load.rb
tenderlove: Yes, I think this is the correct fix.

amomchilov: Opened https://github.com/ruby/psych/pull/568 to handle this

test/psych/test_merge_keys.rb
```diff
@@ -117,8 +117,7 @@ def test_missing_merge_key
 bar:
   << : *foo
       eoyml
-      exp = assert_raise(Psych::BadAlias) { Psych.load yaml }
-      assert_match 'foo', exp.message
+      assert_raise(Psych::AliasesNotEnabled) { Psych.load yaml }
```
amomchilov, line 120: TODO: This passes, but it totally shouldn't.

test/psych/test_merge_keys.rb
```diff
@@ -117,7 +117,7 @@ def test_missing_merge_key
 bar:
   << : *foo
       eoyml
-      exp = assert_raise(Psych::BadAlias) { Psych.load yaml }
+      exp = assert_raise(Psych::BadAlias) { Psych.load(yaml, aliases: true) }
```
amomchilov, line 120: Turns out this test never actually failing because of the `foo` anchor not existing.

Aliases just weren't enabled xD

PR #568, opened 2022-07-22
Clarify tests about parsing aliases

amomchilov:
Split from #567.

These tests don't really have actually anything special to do with recursion. The behaviour being asserted is about how any kind of alias works.

### DIFF
--- a/test/psych/test_array.rb
+++ b/test/psych/test_array.rb
@@ -57,6 +57,19 @@ module Psych
       assert_cycle(@list)
     end
 
+    def test_recursive_array_uses_alias
+      @list << @list
+
+      expected = <<~eoyaml
+        --- &1
+        - :a: b
+        - foo
+        - *1
+      eoyaml
+
+      assert_equal expected, Psych.dump(@list)
+    end
+
     def test_cycle
       assert_cycle(@list)
     end
--- a/test/psych/test_array.rb
+++ b/test/psych/test_array.rb
@@ -57,17 +57,20 @@ module Psych
       assert_cycle(@list)
     end
 
+    def test_recursive_array
+      @list << @list
+
+      loaded = Psych.load(Psych.dump(@list), aliases: true)
+
+      assert_same loaded, loaded.last
+    end
+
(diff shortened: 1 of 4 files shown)


tenderlove:
> These tests don't really have actually anything special to do with recursion.

This isn't 100% true.  The test implicitly asserted that `Psych.dump([].tap { |x| x << x })` would emit an alias.  We know that an alias was emitted because we explicitly did not allow loading aliases and an exception was raised.

Can you make sure that we assert that `Psych.dump([].tap { |x| x << x })` will use aliases?

amomchilov:
> Can you make sure that we assert that `Psych.dump([].tap { |x| x << x })` will use aliases?

Good point. I'll make a test that ensures this, explicitly.

amomchilov:
> Can you make sure that we assert that `Psych.dump([].tap { |x| x << x })` will use aliases?

Good point. I added tests that ensures this, explicitly, not just for arrays, but also for hashes and object.

test/psych/test_array.rb
```diff
@@ -57,6 +57,19 @@ def test_self_referential
...
     end
 
+    def test_recursive_array_uses_alias
+      @list << @list
+
+      expected = <<~eoyaml
+        --- &1
+        - :a: b
+        - foo
+        - *1
```
tenderlove, line 67: I don't think we should hard-code the expected YAML.  We may want to change the emitted aliases in the future, and now this test couples the implementation with the exact alias name.  IOW I want to test that aliases _were used_, but I don't care what the aliases were.

Not to mention, we may change indentation in the future etc, so I don't want the test to be coupled to the formatting either.

tenderlove (changes requested):

test/psych/test_array.rb
amomchilov: Done, what do you think? [`26af85e` (#568)](https://github.com/ruby/psych/pull/568/commits/26af85e25343c7d1e35d10ba80ceab6efb1cf85c)

Issue #572, opened 2022-08-08
Calendar info in `Date`/`DateTime` is lost


nobu:
```sh-session
$ ruby -rdate -rpsych \
  -e 'di = Date.new(1000, 1, 1, Date::ITALY)' \
  -e 'dg = Date.new(1000, 1, 1, Date::GREGORIAN)' \
  -e 'puts Psych.dump([di, dg]), di == dg'
---
- 1000-01-01
- 1000-01-01
false
```

As shown by the last `false` `di` and `dg` are different dates, but cannot distinguish in the YAML dumped by Psych.

PR #573, opened 2022-08-09
Dump Date/DateTime as proleptic Gregorian date as well as Time

nobu:
Fix ruby/psych#572


nobu:
The [YAML spec] does not seem to mention about calendar systems at all.

[YAML spec]: https://yaml.org/spec/

nobu:
Asked yaml/yaml#69.

v2kovac:
I'm pretty sure this broke my deserialization of Date.

```
a  = Date.today.to_yml # "--- 2023-06-07\n"
YAML.unsafe_load(a) # return "2023-06-07" instead of Date object
```

very confusing since this is a minor version change, i don't see other people complaining about this though

PR #592, opened 2022-10-12
Abandon when libyaml is not found

nobu:


ext/psych/extconf.rb
```diff
@@ -44,6 +36,11 @@
   libyaml = "libyaml.#$LIBEXT"
   $cleanfiles << libyaml
   $LOCAL_LIBS.prepend("$(LIBYAML) ")
+else # default to pre-installed libyaml
+  pkg_config('yaml-0.1')
+  dir_config('libyaml')
+  find_header('yaml.h') or abort "yaml.h not found"
+  find_library('yaml', 'yaml_get_version') or "libyaml not found"
```
olleolleolle, line 43: Q: Is there a missing `abort` here?

```suggestion
  find_library('yaml', 'yaml_get_version') or abort "libyaml not found"
```

nobu: Indeed.

hsbt (approved):

PR #601, opened 2022-12-16
Add msys2 dependencies for windows

dsisnero:


hsbt:
Thanks. Can you point the specification of `msys2_mingw_dependencies` ?

I' not sure what this is correct declaration for dependencies. Should we add them with `Array` like `['libyaml', 'libyaml-devel']`?

dsisnero:
https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers

Their example
gemspec.metadata['msys2_mingw_dependencies'] = 'libusb sqlite'

MSP-Greg:
Most MSYS2 mingw/ucrt packages do not have 'dev' or 'devel' packages, unlike many *nix derivates.  Specifically, there is no `libyaml-devel` package.  Otherwise, LGTM.

PR #610, opened 2023-01-11
Get rid of anonymous eval calls

casperisfine:
Things declared in anonymous eval are always annoying to locate.


casperisfine:
cc @hsbt @tenderlove

hsbt:
It's reasonable. Thanks!

PR #612, opened 2023-01-13
Move JRuby extension to SnakeYAML Engine

headius:
See jruby/jruby#7570 for some of the justification for this move. We only require the parser from SnakeYAML, but in the original form it is encumbered with Java object serialization code that keeps getting flagged as a CVE risk. We disagree with the assessment, at least as it pertains to JRuby (we do not use the code in question) but our inclusion of the library continues to get flagged by auditing tools.

This PR starts the process of moving to the successor library, SnakeYAML Engine. The parser API is largely unchanged, except as seen in this commit. No Java exceptions are thrown, but a number of Psych tests fail (possibly due to Engine being YAML 1.2 only).

Issue #579, opened 2022-09-11
JRuby: Provide some sort of ability to configure SnakeYAML LoaderOptions


chadlwilson:
Currently, it appears the JRuby Psych extension does not configure the SnakeYaml [`LoaderOptions`](https://bitbucket.org/snakeyaml/snakeyaml/src/master/src/main/java/org/yaml/snakeyaml/LoaderOptions.java) or allow the user to override them somehow.

The parser seems to be constructed at 
https://github.com/ruby/psych/blob/a565e1fcec3dad4de7e753b9c3e113c7ed1a22a4/ext/java/org/jruby/ext/psych/PsychParser.java#L195-L201

There is an [optional second arg](https://bitbucket.org/snakeyaml/snakeyaml/src/261c60857a64927efd53112fe53072ba6ae5decc/src/main/java/org/yaml/snakeyaml/parser/ParserImpl.java#lines-144) to allow specifying options.

There is some discussion related to this because of CVEs being raised by OSS Fuzz about the risks of using SnakeYaml (or most YAML parsers really) with fully untrusted content. I am unclear if folks might be using JRuby and thus Psych to do this, e.g a YAML-based API or config approach. Having said this, the default `LoaderOptions` do seem to be relatively "secure by default" so this doesn't seem to be a security issue.

Possible things folks might want to do with `LoaderOptions`.
1. Increase/decrease `maxAliasesForCollections` or `nestingDepthLimit` which are DoS/billion laughs mitigations
2. Disallow duplicate keys `allowDuplicateKeys`

Related discussion
- https://github.com/jruby/jruby/issues/7342#issuecomment-1242888995
- https://github.com/ruby/psych/pull/426#issuecomment-625597776
- https://github.com/ruby/psych/pull/574

headius:
This is a good idea. I'm not sure the best way to include it without introducing an incompatibility. Does libyaml have such configuration options? It would be best if we could provide a uniform interface for setting options, and if possible map the same options to the same logic in both backend libraries.

chadlwilson:
Thanks @headius - that looks like a reasonable solution to me. Thanks for your efforts here and looking forward to JRuby `9.4.1.0`! 😀

chadlwilson:
[...]

PR #613, opened 2023-01-13
Expose load settings

headius:
This builds off #612, adding access to some key DOS-related settings in the parser. See #579.

The PR in #612 should be merged before this one (and I will rebase to master when that happens).

### DIFF
--- a/ext/java/org/jruby/ext/psych/PsychParser.java
+++ b/ext/java/org/jruby/ext/psych/PsychParser.java
@@ -34,6 +34,7 @@ import org.jcodings.specific.UTF8Encoding;
 import org.jcodings.unicode.UnicodeEncoding;
 import org.jruby.Ruby;
 import org.jruby.RubyArray;
+import org.jruby.RubyBoolean;
 import org.jruby.RubyClass;
 import org.jruby.RubyEncoding;
 import org.jruby.RubyFixnum;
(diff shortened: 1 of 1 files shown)


headius:
@hsbt @tenderlove I would like to start shipping this feature, so users can configure the limits in the SnakeYAML parser. Is what I have here acceptable? Does libyaml have any similar configurations we could expose in the same way?

hsbt:
It's reasonable to expose `CodePointLimit` for https://github.com/jruby/jruby/issues/7543 and `AllowDuplicateKeys` for https://github.com/ruby/psych/pull/426#issuecomment-625597776

But I'm not sure we should allow to set `AllowRecursiveKeys` and `MaxAliasesForCollections`. Can you point why we should allow them?

headius:
@hsbt I believe all four of these were requested by users. Those additional two (recursive keys, max aliases for collections) were mentioned in #579.

hsbt:
Ah, I see. Thanks.

hsbt (approved):

headius:
I guess this is approved so I'm going to merge. If we find that these settings are configurable in libyaml we can deal with making the API the same at that point.

yvesll:
What about this option in high-level API like Psych.load? Now I need to add a monkey patch to set the code point limit in Psych module method..., but this seems not a bad way.

headius:
@yvesll That's not a bad idea. Perhaps you could put together a PR that allows passing these values through as options?

oliverbarnes:
How are can these load settings used in a rails app? (JRuby newbie here)

I tried implementing the monkey patch as per @yvesll comment, but that isn't working:

```
#Rails initializer
module Psych
  class Parser
    def code_point_limit(_context)
      20_000_000
    end
  end
end
```

We're on Rails 6 running on JRuby 9.4.3.0, and we're hitting the code point limit issue when loading big yaml fixtures for rspec:

```
Failure/Error: Unable to find org.snakeyaml.engine.v2.scanner.ScannerImpl.fetchMoreTokens(ScannerImpl.java to read failed line

     Java::OrgSnakeyamlEngineV2Exceptions::YamlEngineException:
       The incoming YAML document exceeds the limit: 3145728 code points.
     # org.snakeyaml.engine.v2.scanner.ScannerImpl.fetchMoreTokens(ScannerImpl.java:289)
     # org.snakeyaml.engine.v2.scanner.ScannerImpl.checkToken(ScannerImpl.java:192)
     # org.snakeyaml.engine.v2.parser.ParserImpl$ParseBlockMappingKey.produce(ParserImpl.java:710)
     # org.snakeyaml.engine.v2.parser.ParserImpl.lambda$produce$1(ParserImpl.java:232)
     # java.base/java.util.Optional.ifPresent(Optional.java:178)
     # org.snakeyaml.engine.v2.parser.ParserImpl.produce(ParserImpl.java:232)
     # org.snakeyaml.engine.v2.parser.ParserImpl.hasNext(ParserImpl.java:226)
     # org.jruby.ext.psych.PsychParser.parse(PsychParser.java:245)
     # org.jruby.ext.psych.PsychParser$INVOKER$i$3$0$parse.call(PsychParser$INVOKER$i$3$0$parse.gen)
```

Is there a way to set it in a .jrubyrc config? I didn't find a `code_point_limit` option in `jruby --properties` output

yvesll:
HI, @oliverbarnes, here is my monkey patch which works fine for me, hope it can help you:
```ruby
module Psych
  def self.parse_stream yaml, filename: nil, &block
    if block_given?
      parser = Psych::Parser.new(Handlers::DocumentStream.new(&block))
      parser.code_point_limit = 20_000_000
      parser.parse yaml, filename
    else
      parser = self.parser
      parser.parse yaml, filename
      parser.code_point_limit = 20_000_000
      parser.handler.root
    end
  end if defined?(RUBY_ENGINE) && RUBY_ENGINE == "jruby"
end
```

oliverbarnes:
Thanks @yvesll! I'm getting a new error now, but I think it's enough to make progress:

```
the leading empty lines contain more spaces (8) than the first non-empty line. while scanning a block scalar at line 32385 column 5
     # ./config/initializers/psych.rb:6:in `parse_stream'
```

headius:
I had not considered that many applications won't control the construction of the psych parser! I think we should add class methods to psych for setting the default values that then will be passed in for every parser creation. It may also be worth using higher defaults to begin with. Perhaps one of you could try to create a PR for that feature?

oliverbarnes:
@headius I'd be happy to do that (after some grokking), but I'm still encountering issues related to psych.

Using @yvesll's monkey patch got me past the code point limitation, but now I'm stuck with VCR not recognizing http requests previously recorded in yaml fixtures. I have a hunch this is due to Psych now being compliant with YAML 1.2, as mentioned in other issues here on the repo.

I'm spiking jruby for a client, and before I can dedicate time contributing to the former, I need to show some progress on it :)

headius:
> due to Psych now being compliant with YAML 1.2

@oliverbarnes That is certainly possible, but there's always the possibility of a bug in SnakeYAML Engine. If you can show this parsing with another 1.2-compliant parser, we will know.

Keep me posted...I want to help your experiment to succeed!

headius:
@oliverbarnes Is there an issue filed for your YAML 1.2 problem? I want to make sure we track it but I'm not sure what other issues you were referring to. I found only one recent issue in the psych tracker since the upgrade.

Given your report and the other I found I'm recommending that we not upgrade psych in the JRuby 9.3 line (jruby/jruby#7600), so having a complete bug report would be really helpful!

headius:
@oliverbarnes Here, I'll trade you this for your YAML 1.2 issue report: https://github.com/ruby/psych/pull/647

oliverbarnes:
> Here, I'll trade you this for your YAML 1.2 issue report: https://github.com/ruby/psych/pull/647

:D fair trade! Thanks, and will do @headius 👍 

> Keep me posted...I want to help your experiment to succeed!

For sure! Appreciate the support :)

PR #628, opened 2023-05-10
Ensure strings that may be confused as YAML1.2 numbers are quoted

adammw:
Ensure strings that may be confused as YAML1.2 numbers are quoted.

Example: `0o123` or `1e3` strings in Ruby will be printed without quotation marks, resulting in them changing to numbers in a YAML1.2-parser (like that used in [Go](https://github.com/go-yaml/yaml)).


test/psych/visitors/test_yaml_tree.rb
```diff
@@ -167,9 +167,19 @@ def test_float
...
         assert_match(/'01818'/, Psych.dump({'a' => '01818'}))
+
+        # YAML 1.2 int Base 8
+        assert_match(/'0o17'/, Psych.dump({'a' => '0o17'}))
+        assert_match(/'0o111'/, Psych.dump({'a' => '0o111'}))
+
+        # YAML 1.2 float Number
+        assert_match(/'12e03'/, Psych.dump({'a' => '12e03'}))
+        assert_match(/'1.2e3'/, Psych.dump({'a' => '1.2e3'}))
+        assert_match(/".2e3"/, Psych.dump({'a' => '.2e3'}))
```
nobu, line 182: `.` (any character) does not seem what you want here.

```suggestion
        assert_include(Psych.dump({'a' => '0o17'}), "'0o17'")
        assert_include(Psych.dump({'a' => '0o111'}), "'0o111'")

        # YAML 1.2 float Number
        assert_include(Psych.dump({'a' => '12e03'}), "'12e03'")
        assert_include(Psych.dump({'a' => '1.2e3'}), "'1.2e3'")
        assert_include(Psych.dump({'a' => '.2e3'}), "".2e3"")
```

djhoese:
@adammw Any chance you have time to continue working on this? I've run into this same problem where strings (git commit SHAs) that look like scientific notation are dumped with no quotes around them.

adammw:
@djhoese Ironically that's the exact reason I started looking into this 😅 but also why it's been pretty low priority, the "solution" is just to rewrite the commit to get a different SHA so it doesn't hit this condition

test/psych/visitors/test_yaml_tree.rb
adammw: Updated

CarlosCD:
This seems to solve #701  as well 

❤️

PR #633, opened 2023-06-12
Update spec files

nobu:


psych.gemspec
```diff
@@ -21,22 +21,35 @@ DESCRIPTION
   s.licenses = ["MIT"]
   s.require_paths = ["lib"]
 
-  # for ruby core repository. It was generated by `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+  # for ruby core repository.
+  # It was generated by
+  `git ls-files -z`.split("\x0").reject { |f|
```
hsbt, line 26: @nobu This line was invoked. Is it intentional?

Issue #649, opened 2023-09-19
Add settings for alias limits, recursion, and duplicate keys


headius:
In #613 and #647 I exposed settings in SnakeYAML Engine for limiting code points, aliases, recursive keys, and duplicate keys.

I was able to add one test that worked, for the code point limit setting. I could not get the others to error, and I think I've figured out why: they are not done at the parser level.

The maximum alias detection, for example, which is used to prevent the "billion laughs" attack, is done above the parser in SnakeYAML Engine as part of composing YAML nodes for the rest of the library. JRuby's Psych backend bypasses the composer/node API and uses the parser directly, with the node-wrangling logic living in the rest of Psych.

https://bitbucket.org/snakeyaml/snakeyaml-engine/src/8fc70e5d943a24ed9eb94eb333a4da7ca79e62c1/src/main/java/org/snakeyaml/engine/v2/composer/Composer.java#Composer.java-187:194

I suspect this is the same situation for the duplicate and recursive keys, but hopefully @asomov can tell us if this is the case.

If so, then three out of the four settings I added are not really being used by SnakeYAML Engine, and they should be changed into Psych-level settings. However I am not sure how the C version of Psych detects and prevents these situations. Need clarification from @tenderlove or @hsbt.

Alternatively, it may make sense for the SnakeYAML Engine parser itself to honor these settings, but that would be up to @asomov.

headius:
@oliverbarnes Ping! I believe this is why I could not get the other settings to work properly. I'm glad the one you needed is hooked up, though!

headius:
@tenderlove @hsbt I would like your input here. In #647 I added the ability to set the SnakeYAML defaults for four settings, but it appears only one of those four is actually checked in the parser (maximum code points). The others appear to be done as part of SnakeYAML's node graph building, such as the alias limit I mention in this issue's description.

In order for these settings to be useful, one of two things would need to happen:

* SnakeYAML Engine would need to check them during parse, rather than during tree building or other higher-level parts of the API as it does now; OR
* Psych would need to add equivalent checks within its own tree-building etc logic. (Probably the more logical place since it would bring these limits to all Psych users).

Either of these would require work beyond my expertise, so I don't expect them to happen soon. And that leaves us with three JRuby-specific settings that don't really do anything (they do, but they don't affect execution of the parts of SnakeYAML Engine that Psych uses).

[...]

PR #647, opened 2023-09-10
Allow setting defaults for SnakeYAML limits

headius:
This adds class-level accessors for the following SnakeYAML-specific parser settings:

* max_aliases_for_collections
* allow_duplicate_keys
* allow_recursive_keys
* code_point_limit

The initial values are based on SnakeYAML Engine's defaults. This PR does not modify those default values.

Using these accessors, it should be possible for users to globally change them for all future Psych parser instances without resorting to monkey patches as described in https://github.com/ruby/psych/pull/613#issuecomment-1707712390.

PR #653, opened 2023-10-05
Remove SnakeYAML settings that don't work

headius:
These settings are not enforced at the parser level in SnakeYAML Engine, instead being enforced during the construction of the YAML node graph. Because the JRuby Psych wrapper around SnakeYAML only uses the parser directly, the settings have no effect. This commit removes the ineffective settings until we can decide what to do about them.

See #613, #647, and #649.


headius:
@hsbt This is merged and could be released. Let me know if you will do it or if I should do it.

headius:
@tenderlove @hsbt Can we release this week? JRuby 9.4.4.0 should be next week and I'd like to get this update in there.

hsbt:
@headius I released Psych 5.1.1 with this fix. Let me know if you have another request.

PR #663, opened 2023-12-18
`YAMLTree` performance by not using `object_id`s

amomchilov:
Since Ruby 2.7, `object_id` became more expensive, particularly on its first call for a particular object. @JemmaIssroff has a great [blog post](https://jemma.dev/blog/gc-object-id) about it.

We can reduce how many objects we assign object IDs to, by using `Hash#compare_by_identity` in the `YAMLTree::Registrar`. The semantics will be the same, but now loads/stores into the Hash are faster, and we don't need to trigger lots of object ID assignments.

On a related note, I used `assert_same` in the tests, where applicable (instead of `assert_equal a.object_id, b.object_id`).

Is there a benchmark suite I can run to compare before/after performance?

### DIFF
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -15,30 +15,29 @@ module Psych
     class YAMLTree < Psych::Visitors::Visitor
       class Registrar # :nodoc:
         def initialize
-          @obj_to_id   = {}
-          @obj_to_node = {}
+          @obj_to_id   = {}.compare_by_identity
+          @obj_to_node = {}.compare_by_identity
           @targets     = []
           @counter     = 0
         end
 
         def register target, node
-          return unless target.respond_to? :object_id
           @targets << target
-          @obj_to_node[target.object_id] = node
+          @obj_to_node[target] = node
         end
 
         def key? target
-          @obj_to_node.key? target.object_id
+          @obj_to_node.key? target
         rescue NoMethodError
           false
         end
 
         def id_for target
-          @obj_to_id[target.object_id] ||= (@counter += 1)
+          @obj_to_id[target] ||= (@counter += 1)
         end
 
         def node_for target
-          @obj_to_node[target.object_id]
+          @obj_to_node[target]
         end
       end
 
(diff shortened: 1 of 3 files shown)


tenderlove:
Great find, thank you!

lib/psych/visitors/yaml_tree.rb
```diff
@@ -15,30 +15,29 @@ module Visitors
     class YAMLTree < Psych::Visitors::Visitor
       class Registrar # :nodoc:
         def initialize
-          @obj_to_id   = {}
-          @obj_to_node = {}
+          @obj_to_id   = {}.compare_by_identity
+          @obj_to_node = {}.compare_by_identity
           @targets     = []
```
amomchilov, line 20: @tenderlove Looking again... is `@targets` just completely unused? Only ever shoved into, never read.

amomchilov: Removed in #665

amomchilov:
Put this tiny synthetic benchmark together and woah... 7% improvement on net serialization performance!

```
Comparison:
  After removing @targets array:     3030.9 i/s
After switching to identity set:     2975.9 i/s - same-ish: difference falls within error
             Before all changes:     2828.5 i/s - 1.07x  slower
```

<details><summary><code>benchmark.rb</code></summary>

```ruby
def generate_test_data
  repeated_points = 10.times.map { |i| [i, i] }
  (repeated_points * 2) + 80.times.map { |i| [i, i] }
end

def test_perf
  puts Psych.dump(generate_test_data)

  points = generate_test_data
  Benchmark.ips do |x|
    x.report("             Before all changes") do |times|
      i = 0
      while (i += 1) < times
        Psych.dump(points.map { |p| p.dup })
      end
    end

    x.hold!("psych_benchmark1.json")
    # Then: git checkout 0dc25a9

    x.report("After switching to identity set") do |times|
      i = 0
      while (i += 1) < times
        Psych.dump(points.map { |p| p.dup })
      end
    end

    x.hold!("psych_benchmark2.json")
    # Then: git checkout 6905a2

    x.report("  After removing @targets array") do |times|
      i = 0
      while (i += 1) < times
        Psych.dump(points.map { |p| p.dup })
      end
    end

    x.compare!
  end
end
```
</details>

The difference is emphasized here, I chose objects that do minimal other serialization (e.g. simple arrays, no custom class with long names, and cheap field values).

PR #665, opened 2023-12-19
Remove unused array

amomchilov:
I noticed in #663 that the `@targets` ivar in `Psych::YAMLTree::Registrar` is only ever written into, and never read.

### DIFF
--- a/lib/psych/visitors/yaml_tree.rb
+++ b/lib/psych/visitors/yaml_tree.rb
@@ -26,8 +26,6 @@ module Psych
 
         def key? target
           @obj_to_node.key? target
-        rescue NoMethodError
-          false
         end
 
         def id_for target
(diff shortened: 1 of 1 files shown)


lib/psych/visitors/yaml_tree.rb
```diff
@@ -17,19 +17,15 @@ class Registrar # :nodoc:
...
 
         def register target, node
-          @targets << target
           @obj_to_node[target] = node
         end
 
         def key? target
           @obj_to_node.key? target
-        rescue NoMethodError
-          false
```
amomchilov, line 32: I missed this in #663.

I believe it existed to just catch if `object_id` was not defined, but since we don't call that anymore, we don't have to worry about it being missing.

tenderlove:
Excellent! Thank you!

PR #670, opened 2024-04-25
instructions for macports users

jjb:
This is probably not the best place to put this

and/or we can just add brew as well?

for homebrew, are the flags needed, or does it look in those paths by default?

can/should the project look into the macports paths by default as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,11 @@ Psych supported the static build with specific version of libyaml sources. You c
 
     gem install psych -- --with-libyaml-source-dir=/path/to/libyaml-0.2.5
 
+For MacPorts users:
+
+    sudo port install yaml-cpp libyaml
+    gem install psych -- --with-opt-include=/opt/local/include --with-opt-lib=/opt/local/lib
+
 In order to use the gem release in your app, and not the stdlib version,
 you'll need the following:
 


tenderlove:
Homebrew provides pkg-config files for libyaml, so I think that's how [Psych picks it up](https://github.com/ruby/psych/blob/be0ba74e5613c20f213403e15914d24944c2652d/ext/psych/extconf.rb#L40-L41).

Are you able to do this?

```
$ pkg-config --list-all | grep yaml
yaml-0.1          LibYAML - Library to parse and emit YAML
```

jjb:
```
pkg-config --list-all | grep yaml
```

jjb:
```
sudo port install pkgconfig

## open new terminal.....

➔ pkg-config --list-all | grep yaml
yaml-cpp        Yaml-cpp - A YAML parser and emitter for C++
yaml-0.1        LibYAML - Library to parse and emit YAML
```

jjb:
```
sudo port uninstall pkgconfig
mv .bundle/config .bundle/config-back
gem uninstall psych
## open new window
gem install psych
# it breaks!
sudo port install pkgconfig
## open new terminal
gem install psych
# it works!
```

ARE YOU KIDDING ME
HOW HAVE I BEEN PROGRAMMING FOR 23 YEARS AND DIDN'T KNOW ABOUT THIS

PR #687, opened 2024-09-09
Ensure strings with only underscores are not processed as Integer

zendesk-jmeade:
A string similar to "0x____" should be treated as a string. Currently it is processed as an Integer.

This alters the regex specified by http://yaml.org/type/int.html to ensure at least one numerical symbol is present in the string before converting to Integer.

### DIFF
--- a/lib/psych/scalar_scanner.rb
+++ b/lib/psych/scalar_scanner.rb
@@ -13,18 +13,18 @@ module Psych
     # Base 60, [-+]inf and NaN are handled separately
     FLOAT = /^(?:[-+]?([0-9][0-9_,]*)?\.[0-9]*([eE][-+][0-9]+)?(?# base 10))$/x
 
-    # Taken from http://yaml.org/type/int.html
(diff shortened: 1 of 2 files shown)


KJTsanaktsidis (approved):
❤️ 

I think the _spec_ is also broken here. As you say, these regexes came from the spec.

```
---
foo: 0x_
```

Since the spec says that the regexes in question "may also be used for implicit tag resolution", that means that the plain-style `0x_` is expected to become an integer.... but... there's really no way to say _what_ integer that's supposed to be 😅 

The YAML 1.2 spec though actually deals with this issue already by not allowing `_` in integer literals - https://yaml.org/spec/1.2.2/#1012-tag-resolution. So actually this is _only_ a 1.1 issue. I don't know if there's a process for making errata for YAML 1.1.

I opened an issue about the spec here: https://github.com/yaml/yaml-spec/issues/338

KJTsanaktsidis:
cc @tenderlove @hsbt this change looks good to me 👍 but you're the owners of the gem so I'll leave merging & releasing up to you when you're happy to do it. Thanks!

Issue #694, opened 2024-11-13
Breaking change in 5.2.0


tdeo:
I've run into the following behavior after upgrading to psych v5.2.0:

#### With 5.1.2

```ruby
irb(main):001> require 'yaml'
=> true
irb(main):002> Gem.loaded_specs['psych'].version
=> Gem::Version.new("5.1.2")
irb(main):003> '2020-01-01'.to_yaml
=> "--- '2020-01-01'\n"
```

Observe how the string is quoted in the Yaml file

#### With 5.2.0

```ruby
irb(main):001> require 'yaml'
=> true
irb(main):002> Gem.loaded_specs['psych'].version
=> Gem::Version.new("5.2.0")
irb(main):003> '2020-01-01'.to_yaml
=> "--- 2020-01-01\n"
```

Now the string is unquoted, which makes it invalid to be parsed back with `YAML.safe_load`.

I believe this behavior originates from [this commit](https://github.com/ruby/psych/commit/06db36fa3a849f229e2cbc7b495d142ff58a1f89), also because I noticed that manually requiring `'date'` before calling `.to_yaml` still yields the expected result of `v5.1.2`:

```ruby
irb(main):001> require 'yaml'
=> true
irb(main):002> Gem.loaded_specs['psych'].version
=> Gem::Version.new("5.2.0")
irb(main):003> require 'date'
=> true
irb(main):004> '2020-01-01'.to_yaml
=> "--- '2020-01-01'\n"
```

I think the following error probably hints as to where the problem originates from:
```irb
irb(main):001> require 'yaml'; Psych::ClassLoader.new.date
/Users/thierry/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/psych-5.2.0/lib/psych/class_loader.rb:58:in `path2class': undefined class/module Date (ArgumentError)

        path2class(name)
                   ^^^^
	from /Users/thierry/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/psych-5.2.0/lib/psych/class_loader.rb:58:in `resolve'
	from /Users/thierry/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/psych-5.2.0/lib/psych/class_loader.rb:50:in `find'
	from /Users/thierry/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/psych-5.2.0/lib/psych/class_loader.rb:30:in `load'
[...]

PR #695, opened 2024-11-13
Eagerly require `date`

tdeo:
Fixes https://github.com/ruby/psych/issues/694. 

When picking up only the test changes (and not `require 'date'` explicitly anymore), such failures start to appear:
```
Error: test_spec_sequence_key_shortcut(Psych_Unit_Tests): NameError: uninitialized constant Psych_Unit_Tests::Date
/Users/thierry/workspace_personal/psych/test/psych/test_yaml.rb:299:in `test_spec_sequence_key_shortcut'
     296:     def test_spec_sequence_key_shortcut
     297:         # Shortcut sequence map
     298:         assert_parse_only(
  => 299:           { 'invoice' => 34843, 'date' => Date.new( 2001, 1, 23 ),
     300:             'bill-to' => 'Chris Dumars', 'product' =>
     301:             [ { 'item' => 'Super Hoop', 'quantity' => 1 },
     302:               { 'item' => 'Basketball', 'quantity' => 4 },
=======================================================================================================================
Finished in 0.390079 seconds.
-----------------------------------------------------------------------------------------------------------------------
609 tests, 1481 assertions, 0 failures, 21 errors, 0 pendings, 0 omissions, 0 notifications
96.5517% passed
```

The `autoload :Date` statement gets moved to a more appropriate file (the one actually loading the constant), and moved to the top-level so it requires `::Date` and not try to require `Psych::ClassLoader::Date`

Performance-wise, reusing the benchmark from https://github.com/ruby/psych/commit/06db36fa3a849f229e2cbc7b495d142ff58a1f89#diff-6a459e056cadf37665f54005bd2dde09d9ba8e66c9807eb0dc67145f9b841771R7, it seems to further improve performance quite unsignificantly:

With those changes:
```
$ ruby /tmp/bench-yaml.rb
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin23]
           100 dates   464.000 i/100ms
           100 dates      4.605k (± 2.3%) i/s  (217.18 μs/i) -     23.200k in   5.041382s
```

On current master:
```
$ ruby /tmp/bench-yaml.rb
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin23]
           100 dates   439.000 i/100ms
           100 dates      4.515k (± 1.9%) i/s  (221.49 μs/i) -     22.828k in   5.058050s
```

### DIFF
--- a/psych.gemspec
+++ b/psych.gemspec
@@ -75,6 +75,8 @@ DESCRIPTION
     s.add_dependency 'stringio'
   end
 
+  s.add_dependency 'date'
+
   s.metadata['msys2_mingw_dependencies'] = 'libyaml'
   s.metadata['changelog_uri'] = s.homepage + '/releases'
 end
--- a/test/psych/test_string.rb
+++ b/test/psych/test_string.rb
@@ -50,6 +50,16 @@ module Psych
       assert_equal str, Psych.load(yaml)
     end
 
+    def test_single_quote_when_matching_date
+      pend "Failing on JRuby" if RUBY_PLATFORM =~ /java/
+
+      lib = File.expand_path("../../../lib", __FILE__)
+      assert_separately(["-I", lib, "-r", "psych"], __FILE__, __LINE__ + 1, <<~'RUBY')
+        yml = Psych.dump('2024-11-19')
+        assert_equal '2024-11-19', Psych.load(yml)
+      RUBY
+    end
+
     def test_plain_when_shorten_than_line_width_and_no_final_line_break
       str = "Lorem ipsum"
       yaml = Psych.dump str, line_width: 12
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'date'
+
 require_relative 'psych/versions'
 case RUBY_ENGINE
 when 'jruby'
--- a/lib/psych/scalar_scanner.rb
+++ b/lib/psych/scalar_scanner.rb
@@ -4,8 +4,6 @@ module Psych
   ###
   # Scan scalars for built in types
   class ScalarScanner
-    autoload :Date, "date"
-
     # Taken from http://yaml.org/type/timestamp.html
     TIME = /^-?\d{4}-\d{1,2}-\d{1,2}(?:[Tt]|\s+)\d{1,2}:\d\d:\d\d(?:\.\d*)?(?:\s*(?:Z|[-+]\d{1,2}:?(?:\d\d)?))?$/
 
--- a/lib/psych/visitors/to_ruby.rb
+++ b/lib/psych/visitors/to_ruby.rb
@@ -79,7 +79,6 @@ module Psych
           class_loader.big_decimal._load o.value
         when "!ruby/object:DateTime"
           class_loader.date_time
-          require 'date' unless defined? DateTime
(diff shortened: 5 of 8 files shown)


lib/psych/class_loader.rb
```diff
@@ -2,6 +2,8 @@
 require_relative 'omap'
 require_relative 'set'
 
+autoload :Date, "date"
```
byroot, line 5: I don't think we should set consts on Object like that.

byroot (commented):
I don't see any test, so it's hard to confirm it's indeed the cause.

byroot:
You can probably use a subprocess.

lib/psych/class_loader.rb
tdeo: Would you rather change it to a `require 'date'` statement at the top-level?

tdeo:
> You can probably use a subprocess.

I was able indeed to add a test case with this technique, thanks for the suggestion

test/psych/test_string.rb
```diff
@@ -223,6 +223,13 @@ def test_string_with_ivars
       assert_equal ivar, food.instance_variable_get(:@we_built_this_city)
     end
 
+    def test_string_matching_date
+      assert_separately(%w[-r ./lib/psych], <<~RUBY)
```
byroot, line 227:
```suggestion
      assert_separately(%w[-I lib -r psych], <<~RUBY)
```

I think you need to pass `-I` otherwise it risks loading the extension from stdlib.

lib/psych/class_loader.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require_relative 'omap'
 require_relative 'set'
+require 'date'
```
byroot, line 4: We probably should declare `date` in the gemspec?

byroot:
I updated your PR.

cc @tenderlove what do you think?

psych.gemspec
```diff
@@ -75,6 +75,8 @@ DESCRIPTION
     s.add_dependency 'stringio'
   end
 
+  s.add_dependency 'date'
```
tdeo, line 78: Does that make sense for the stdlib?

tdeo:
Thanks for wrapping it up

tenderlove (commented):
Why do we need to do this?  The change seems fine, I'm just not following why.

byroot:
Right, sorry Aaron, I should have explained.

In `YAMLTree#visit_String`, there's this condition to figure if a string should be quoted:

```ruby
        elsif not String === @ss.tokenize(o) or /\A0[0-7]*[89]/.match?(o)
          style = Nodes::Scalar::SINGLE_QUOTED
```

So we end up in:

```ruby
class_loader.date.strptime(string, '%F', Date::GREGORIAN)
```

Which further end up trying `ClassLoader#resolve` of "Date", and this uses `path2class`, so it's resolved as essentially `::Date`, so the scoped `autoload` I introduced isn't triggered in this case.

Hence why the initial version of this PR was to move the autoload at the top level, but at that point I think we should just eager load.

tenderlove:
@byroot ah, gotcha.  I somewhat worry this might have downstream effects (maybe on Bundler or RubyGems).  But I like the change so I'll merge it.

PR #715, opened 2025-02-04
Fix loading/parsing regular expressions with forward slashes

### LINKED ISSUE #523: Dump/Load regexp with escaped slash
### Problem

`$ irb -r psych`

```ruby
irb(main):001:0> Psych.load(Psych.dump(/\d/)) == /\d/
=> true

irb(main):002:0> Psych.load(Psych.dump(/\//)) == /\//
=> false
```

## STR

```
$ cat  psych_regexp.rb
```

```ruby
# frozen_string_literal: true

require 'psych'
require 'test/unit'

class TestPyschRegexpDumpLoad < Test::Unit::TestCase
   def test_psych_version
     assert '3.1.0', Psych::VERSION
   end

   def test_dump_load_regexp_no_escape
     assert_equal "--- !ruby/regexp /^[0-9]+/\n", Psych.dump(/^[0-9]+/)
     assert_equal /^[0-9]+/, Psych.load(Psych.dump(/^[0-9]+/))
   end

   def test_dump_load_regexp_with_slash
     assert_equal "--- !ruby/regexp /^\\//\n", Psych.dump(/^\//)
     assert_equal /^\//, Psych.load(Psych.dump(/^\//))
   end
end

```

## Run

```
$ ruby psych_regexp.rb
```

## Actual Result
```
Loaded suite psych_regexp
Started
.F
=====================================================================================================================================================================================
     15:
     16:    def test_dump_load_regexp_with_slash
     17:      assert_equal "--- !ruby/regexp /^\\//\n", Psych.dump(/^\//)
  => 18:      assert_equal /^\//, Psych.load(Psych.dump(/^\//))
     19:    end
     20: end
psych_regexp.rb:18:in `test_dump_load_regexp_with_slash'
</^\//> expected but was
</^\//>

diff:
  /^\//
Failure: test_dump_load_regexp_with_slash(TestPyschRegexpDumpLoad)
=====================================================================================================================================================================================
.
Finished in 0.006877 seconds.
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
3 tests, 5 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
66.6667% passed
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
436.24 tests/s, 727.06 assertions/s
```

## Expected Result

Test passed

mamhoff:
This fixes the issue where regular expression would come back slightly different after going through a YAML load/dump cycle. Because we're used to having to escape forward slashes in regular expression literals (because the literal is delimited by slashes), but the deserializer takes the literal output from `Regexp#inspect` and feeds it as a string into `Regexp.new`, which expects a string, not a Regexp literal, cycling did not properly work before this commit.

I've also changed the code to be a bit more readable, I hope this doesn't affect performance.

Fixes #523

### DIFF
--- a/lib/psych/visitors/to_ruby.rb
+++ b/lib/psych/visitors/to_ruby.rb
@@ -96,11 +96,11 @@ module Psych
           Float(@ss.tokenize(o.value))
         when "!ruby/regexp"
           klass = class_loader.regexp
-          o.value =~ /^\/(.*)\/([mixn]*)$/m
-          source  = $1
+          matches = /^\/(?<string>.*)\/(?<options>[mixn]*)$/m.match(o.value)
+          source  = matches[:string].gsub('\/', '/')
           options = 0
           lang    = nil
-          $2&.each_char do |option|
+          matches[:options].each_char do |option|
             case option
             when 'x' then options |= Regexp::EXTENDED
             when 'i' then options |= Regexp::IGNORECASE
--- a/test/psych/test_yaml.rb
+++ b/test/psych/test_yaml.rb
@@ -35,6 +35,10 @@ class Psych_Unit_Tests < Psych::TestCase
         assert_cycle(Regexp.new("foo\nbar"))
     end
 
+    def test_regexp_with_slash
+      assert_cycle(Regexp.new('/'))
+    end
+
     # [ruby-core:34969]
     def test_regexp_with_n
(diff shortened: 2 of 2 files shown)


OpakAlex:
+1

PR #724, opened 2025-04-29
Adding safe_load_stream: a ( nearly ) equivalent to load_stream, but safe

OrenGitHub:
Hi !
It seems that some functions have an equivalent `safe_` versions, when user controlled yaml files are handled. Looking at CVEs like [CVE-2024-42363][1], and its [fix][2], I implemented a minimal `safe_load_stream` that handles most relevant use cases.

From a security perspective, it seems that users might benefit from the existence of such a safe variant.
Would be happy to get feedback, add more tests and the like. Thanks !

[1]: https://nvd.nist.gov/vuln/detail/CVE-2024-42363
[2]: https://github.com/zendesk/samson/pull/4071/files

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -672,7 +672,7 @@ module Psych
     documents = parse_stream(yaml, filename: filename).children.map do |child|
       stream = Psych::Nodes::Stream.new
       stream.children << child
-      safe_load stream.to_yaml, permitted_classes: permitted_classes, aliases: aliases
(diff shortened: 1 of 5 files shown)


OrenGitHub:
Hey Aaron [ @tenderlove ],
Do you think we should move forward with this PR ?
I think it makes sense , and would be happy to get feedback / discuss further
Thanks !

tenderlove:
Hi @OrenGitHub,

Yes, I think this is fine. I was going to merge but got distracted.

tenderlove (approved):

PR #725, opened 2025-04-30
Handle Ruby 3.5 new Set class

casperisfine:
Since `Set` no longer is a regular object class holding a Hash it needs to be specially handled.

FYI: @tenderlove @jeremyevans 
NB: Only added tests for now to show the problem. I think ideally we'd keep the exact same payload so that we don't break people applications.

### DIFF
--- a/lib/psych/core_ext.rb
+++ b/lib/psych/core_ext.rb
@@ -17,3 +17,17 @@ end
 if defined?(::IRB)
   require_relative 'y'
 end
+
+
+# TODO: how best to check for builtin Set?
+if defined?(::Set) && Object.const_source_location(:Set) == ["ruby", 0]
+  class Set
+    def encode_with(coder)
+      coder["hash"] = to_h
+    end
+
+    def init_with(coder)
+      replace(coder["hash"].keys)
+    end
+  end
+end
--- a/test/psych/test_psych_set.rb
+++ b/test/psych/test_psych_set.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+require_relative 'helper'
+
+module Psych
+  class TestPsychSet < TestCase
+    def setup
+      super
+      @set = Psych::Set.new
+      @set['foo'] = 'bar'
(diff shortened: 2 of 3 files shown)


lib/psych/core_ext.rb
```diff
@@ -17,3 +17,17 @@ def to_yaml options = {}
...
+  class Set
+    def encode_with(coder)
+      coder["hash"] = to_h
+    end
+
+    def init_with(coder)
+      replace(coder["hash"].keys)
+    end
+  end
+end
```
casperisfine, line 33: So I added the implementation, now I'm just not sure what is the best way to check for builtin Set. I could check `RUBY_VERSION >= "3.5"` but I don't know if that may cause issues with JRuby / Truffle ruby?

I could check if `Set#initialize` has a source_location, but then what if someone decorated it?

Etc.

tekknolagi: You can check if it has a `hash` ivar

casperisfine: Good point. `if Set.new.instance_variable_get(:@hash)`, but I think JRuby might not like it because they already move Set into a core class a long time ago.

So perhaps I should check with `Pysch.load(Pysch.dump(Set.new([1, 2])))`. I'll sleep on it.

tekknolagi: I guess a broader question is: why not move to a more general-purpose implementation completely? Have the behavior change (looks like a set -> doesn't look like a set) after 3.5. Does inter-version consistency matter?

tekknolagi: Oh, that's an internal implementation

tenderlove (approved):

casperisfine:
Ref: https://github.com/Shopify/bootsnap/issues/498

So since Pysch is a default gem, apps may be on Ruby 2.5 but run an older version of Pysch.

So perhaps this patch isn't correct and we should instead have these methods in ruby itself.

PR #728, opened 2025-05-03
Refine Ruby 3.5 Set support.

byroot:
Use feature testing to detect native Set,
and don't rely on `Set#to_h` which wasn't intended as a public method.

### DIFF
--- a/lib/psych/core_ext.rb
+++ b/lib/psych/core_ext.rb
@@ -18,12 +18,19 @@ if defined?(::IRB)
   require_relative 'y'
 end
 
-
-# TODO: how best to check for builtin Set?
-if defined?(::Set) && Object.const_source_location(:Set) == ["ruby", 0]
+# Up to Ruby 3.4, Set was a regular object and was dumped as such
+# by Pysch.
+# Starting from Ruby 3.5 it's a core class written in C, so we have to implement
+# #encode_with / #init_with to preserve backward compatibility.
+if defined?(::Set) && Set.new.instance_variables.empty?
   class Set
     def encode_with(coder)
-      coder["hash"] = to_h
+      hash = {}
+      each do |m|
+        hash[m] = true
+      end
+      coder["hash"] = hash
+      coder
     end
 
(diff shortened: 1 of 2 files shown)


lib/psych/core_ext.rb
```diff
@@ -18,12 +18,19 @@ def to_yaml options = {}
...
 end
 
-
-# TODO: how best to check for builtin Set?
-if defined?(::Set) && Object.const_source_location(:Set) == ["ruby", 0]
+# Up to Ruby 3.4, Set was a regular object and was dumped as such
+# by Pysch.
+# Starting from Ruby 3.5 it's a C extension, so we have to implement
+# #encode_with / #init_with to preserve backward compatibility.
+if defined?(::Set) && !Psych.dump(Set.new([42])).include?("42")
```
jeremyevans, line 25:
```suggestion
if defined?(::Set) && !Psych.dump(Set[42]).include?("42")
```

Just a little simpler.

lib/psych/core_ext.rb
```diff
@@ -18,12 +18,19 @@ def to_yaml options = {}
   require_relative 'y'
 end
 
-
-# TODO: how best to check for builtin Set?
-if defined?(::Set) && Object.const_source_location(:Set) == ["ruby", 0]
+# Up to Ruby 3.4, Set was a regular object and was dumped as such
+# by Pysch.
+# Starting from Ruby 3.5 it's a C extension, so we have to implement
```
jeremyevans, line 23:
```suggestion
# Starting from Ruby 3.5 it's a core class written in C, so we have to implement
```

jeremyevans (approved):

lib/psych/core_ext.rb
headius: Why not check for @hash ivar?

byroot: I was worried it might not behave correctly on other implementations, but yes, it's probably the simpler

PR #732, opened 2025-05-30
Use the latest stable version for trusted publisher

hsbt:
`3.3` will be EOL at two years later. We should point the latest stable version for the trusted publisher.

### DIFF
--- a/.github/workflows/push_gem.yml
+++ b/.github/workflows/push_gem.yml
@@ -23,7 +23,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


hsbt:
@headius Could you look this JRuby failure?

```
Error: test_ruby_data(Psych_Unit_Tests): NoMethodError: undefined method 'init_struct' for an instance of Psych::Visitors::ToRuby
/home/runner/work/psych/psych/lib/psych/visitors/to_ruby.rb:222:in 'visit_Psych_Nodes_Mapping'
```

Issue #669, opened 2024-01-25
yaml parsing is wrong when key's value has leading colon


subeditara:
replication steps:
```
ruby -ryaml -e "a = YAML.load('servers: ::ffff:10.10.10.1'); puts a; puts a['servers']"
{"servers"=>:":ffff:10.10.10.1"}
:ffff:10.10.10.1
```

expected:
```
{"servers"=>"::ffff:10.10.10.1"}
::ffff:10.10.10.1
```

actual result:
```
{"servers"=>:":ffff:10.10.10.1"}
:ffff:10.10.10.1
```

Environment:
```
ruby --version
ruby 2.7.0p0

gem list yaml 
yaml (0.3.0)
```

Python yaml works as expected:
```
python3.9 -c 'import yaml; a=yaml.safe_load("servers: ::ffff:10.10.10.1"); print(a); print(a["servers"]);'
{'servers': '::ffff:10.10.10.1'}
::ffff:10.10.10.1

```

PR #741, opened 2025-08-05
Add option to disable symbol parsing for ScalarParser

calebstewart:
This PR adds an option to `ScalarParser` to disable the parsing of values containing a leading colon as symbols. This is unexpected behavior when compared with other YAML parsers and when compared with the YAML specification, as symbols are not a concept defined by YAML. I do not expect that we can change the default here as that would break backwards compatibility, so this PR maintains current behavior by leaving the default as `parse_symbols: true`, but exposes a way that users can modify the behavior in order to match what is expected based on the YAML specification.

The option is expected to the `load()` family of methods at the top level as well as to the `ToRuby` visitor and the associated node `to_ruby()` method. In all instances, the default value of the new `parse_symbols` parameter is `true` in order to maintain backwards-compatible behavior. Passing a value of `false` for this parameter will result in values with leading colons being treated as regular strings.

### DIFF
--- a/lib/psych.rb
+++ b/lib/psych.rb
@@ -269,10 +269,10 @@ module Psych
   # YAML documents that are supplied via user input.  Instead, please use the
   # load method or the safe_load method.
   #
-  def self.unsafe_load yaml, filename: nil, fallback: false, symbolize_names: false, freeze: false, strict_integer: false
+  def self.unsafe_load yaml, filename: nil, fallback: false, symbolize_names: false, freeze: false, strict_integer: false, parse_symbols: true
     result = parse(yaml, filename: filename)
     return fallback unless result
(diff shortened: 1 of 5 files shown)


calebstewart:
This PR was created in response to #669. For a concrete example of this functionality, I'll use an equivalent example from that original issue:

```rb
irb(main):001:0> require 'yaml'
=> true
irb(main):002:0> a = YAML.load('servers: ::ffff:10.10.10.1')
=> {"servers"=>:":ffff:10.10.10.1"}
irb(main):003:0> b = YAML.load('servers: ::ffff:10.10.10.1', parse_symbols: false)
=> {"servers"=>"::ffff:10.10.10.1"}
irb(main):004:0>
```

PR #744, opened 2025-09-23
Remove warning by not calling find_library after pkg_config

etiennebarrie:
If `pkg_config` returns a truthy value, it found the library and added it to the global values for the Makefile.

Calling `find_library` after a successful `pkg_config` causes -lyaml to appear twice in the LIBS variable in the resulting Makefile, and causes ld on macOS to emit a warning:

    $ bundle exec rake compile 2>&1 | grep warning
    ld: warning: ignoring duplicate libraries: '-lyaml'


hsbt:
Thanks!

PR #745, opened 2025-09-30
CI: test.yml: test-ibm: Use nproc instead of hard-coded number 2.

junaruga:
This PR is just a small refactoring in the CI's test-ibm job to use the `$(nproc)` instead of hard-coded number `2`. I noticed this was better, and applied it on ruby/openssl repository. So, I want to apply the change to this ruby/psych repository.

https://github.com/ruby/openssl/pull/946/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R44


junaruga:
I merged it because the change is too small and refactoring. I hope you like it.

